PR #12, opened 2024-01-11
SecurityGroup

Ghinwa-Arab:
Adding security group tab for users (fix window expand design)


NoraBitarSM:
next step will be for having a shared component for the security groups window content where we will pass 2 lists and it will do the same work
this shared component will be used in other places for the same purpose but with different data

PR #16, opened 2024-01-12
classes

FatimaSalami:
I still have bugs in combo,s behaviours (2 not binding which are operation and function) --> they are filled from database
and value combo is not being filled upon selecting characteristic..
after debugging this i will delete the console messages you may find.

I pushed anyway so that you can see my progress.


NoraBitarSM:
1. please request a page endpoint for the outer grid 
2. there is an issue with the pagination, it is showing Page: NaN of 1

FatimaSalami:
fixing bug and cleaning code

PR #20, opened 2024-01-13
Mig 40

omarZrayka23:
fixed new in look up of inlineGrid


NoraBitarSM:
search for example for "sms-template3" for any functionId, then select the function template lookup in "Journal Voucher"
The lookup will behave then as if no value was selected

PR #27, opened 2024-01-16
Rem 58

omarZrayka23:
textField reference readonly and mandatory


src/repositories/CurrencyTradingSettingsRepository.js
```diff
@@ -39,5 +39,10 @@ export const CurrencyTradingSettingsRepository = {
     get: service + 'getACT',
     set: service + 'setACT',
     del: service + 'delACT'
+  },
+   Reference: {
+    get : service + 'getDE'
```
NoraBitarSM, line 44: you should add this in the Defaults object with qryDE and set2DE

NoraBitarSM (changes requested):

PR #28, opened 2024-01-16
Mig 41

Ghinwa-Arab:
Add Authorization/Access Control/Security Group and add shared item selector component to it.
Did some fixes in users screen concerning item selector component.


NoraBitarSM (changes requested):

NoraBitarSM:
a video is sent with a bug in the behavior
let me know if it's frontend or backend

PR #29, opened 2024-01-16
Mig 22

FatimaSalami:
strategy screen


src/components/Shared/WindowToolbar.js
```diff
@@ -19,6 +19,11 @@ const WindowToolbar = ({ onSave, onClear, onInfo , disabledSubmit, editMode=fals
         Submit
         </Button>
       )}
+       {onApply && (
+        <Button onClick={onApply} variant='contained' disabled={disabledApply}>
```
NoraBitarSM, line 23: you are using onAppy and disabledApply without sending them as props

NoraBitarSM (changes requested):

src/components/Shared/WindowToolbar.js
FatimaSalami: fixed

PR #33, opened 2024-01-17
test-PR-for-fawzi

NoraBitarSM:
added new line in relation-type index.js


src/pages/relation-types/index.js
```diff
@@ -53,6 +53,7 @@ const RelationTypes = () => {
         setErrorMessage({ message: "YOU DON'T HAVE ACCESS TO THIS SCREEN" })
       }
     }
+
```
NoraBitarSM, line 56: no need for this

PR #46, opened 2024-01-23
Rem 66

FatimaSalami:
Refactoring interface
Please check if the ref field is truncated from top on your laptop 
i found a way to increase the space and solve it, but didn't apply it before making sure if it is necessary on all screens.


src/pages/interface/index.js
```diff
@@ -15,41 +13,52 @@ import GridToolbar from 'src/components/Shared/GridToolbar'
 
 // ** API
 import { RequestsContext } from 'src/providers/RequestsContext'
-import { getNewInterface, populateInterface } from 'src/Models/RemittanceSettings/Interface'
```
NoraBitarSM, line 18: getNewInterface and populateInterface are obsolete
delete the files from Models please

NoraBitarSM (changes requested):

PR #58, opened 2024-01-26
fixLayout+addSavebtn

Ghinwa-Arab:
Table header and submit btn  should be fixed


src/pages/sms-function-templates/index.js
```diff
@@ -175,21 +163,25 @@ const SmsFunctionTemplate = () => {
...
+        .filter(row => row.templateId != null)
+        .map(({ functionId, templateId }) => ({ functionId, templateId }))
+    }
+
+    postRequest({
+      extension: SystemRepository.SMSFunctionTemplate.set,
+      record: JSON.stringify(obj)
+    })
+      .then(res => {
+        getGridData({})
```
NoraBitarSM, line 179: no need to request the api again after saving

src/pages/sms-function-templates/index.js
```diff
@@ -41,9 +38,9 @@ const SmsFunctionTemplate = () => {
   const [access, setAccess] = useState(null)
 
   const _labels = {
```
NoraBitarSM, line 40: please fix the labels like we did after the refactoring

NoraBitarSM (changes requested):

PR #130, opened 2024-02-21
Rem 173

omarZrayka23:
search


src/hooks/resource.js
```diff
@@ -1,21 +1,31 @@
...
   })
 
   return {
     access,
     labels,
-    query
+    query: query,
+    search(query) {
+      setSearchValue(query)
+    }
```
hadichahine, line 28: It's good to add invalidate and clear while we're here.

PR #142, opened 2024-03-01
Rem 82

FatimaSalami:
please inform me when merged to delete old translator labels


NoraBitarSM:
Please let the country combo show ref, name and fl name
and talk to rebecca about the window size and buttons position when maximizing

NoraBitarSM:
please make the display field for currency the 3 fields Ref, name and fl name
and remove the field validation message "This field is required"
having the field in red is enough

PR #158, opened 2024-03-11
Rem 170

Alinadimmokdad:
refactoring fi-account-groups

### DIFF
--- a/src/pages/fi-account-groups/Windows/AccountGroupsWindow.js
+++ b/src/pages/fi-account-groups/Windows/AccountGroupsWindow.js
@@ -1,0 +1,29 @@
+import Window from 'src/components/Shared/Window'
+import AccountGroupsForm from '../forms/AccountGroupsForm'
+
+const AccountGroupsWindow = ({
+  onClose,
+  labels,
+  maxAccess,
+  recordId
+}) => {
+  
+  return (
+    <Window
+      id='AccountGroupsWindow'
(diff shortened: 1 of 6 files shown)


src/pages/fi-account-groups/forms/AccountGroupsForm.js
```diff
@@ -0,0 +1,190 @@
...
+  
+    const formik = useFormik({
+        initialValues,
+        enableReinitialize: true,
+        validateOnChange: true,
+        validationSchema: yup.object({
+          reference: yup.string().required('This field is required'),
+          name: yup.string().required('This field is required'),
+          type: yup.string().required('This field is required'),
+          nraRef: yup.string().required('This field is required'),
```
NoraBitarSM, line 56: number range is optional, remove the NRA fields from here

src/pages/fi-account-groups/Windows/AccountGroupsWindow.js
```diff
@@ -0,0 +1,29 @@
...
+  onClose,
+  labels,
+  maxAccess,
+  recordId
+}) => {
+  
+  return (
+    <Window
+      id='AccountGroupsWindow'
+      Title={labels.accountGroups}
```
NoraBitarSM, line 14: the title of the window should be "Account Group" because we're adding 1 account group with each submit

NoraBitarSM (changes requested):

PR #172, opened 2024-03-26
Rem 200

omarZrayka23:
refactoring


NoraBitarSM:
1. window height is very big
2. fix the height in Amount Range tab and Agent tab (as explained in the video)
3. info button should be enabled in edit mode
4. in Dispersal tab, dispersal column should show the name not the number
5. commission tab in amount range: fix the grid height to be = to the window height. we don't need clear or info button in this window
6. Agent tab: select a dispersal type: it will dissappear.

PR #174, opened 2024-03-28
Rem 198

Alinadimmokdad:
document Release Group Refactoring


NoraBitarSM:
1. add a new record and save, approver tab is still disabled
2. remove the pagination toolbar from the approver tab 
3. approver table should have 2 columns, reference and name. 
4. expand the window's height and width
5. remove the info button from the approver window

NoraBitarSM:
Screenshot 2024-03-29 083415
1. check the table height in approver tab
2. approver table should have 2 columns, reference and name.

PR #198, opened 2024-04-19
Rem 205

RebeccaFrem02:


src/components/Shared/GlobalIntegrationGrid/index.js
```diff
@@ -0,0 +1,178 @@
...
+      getPostTypes();
+    }, []); 
+
+    const getPostTypes = () => {
+      getRequest({
+        extension: GeneralLedgerRepository.IntegrationPostTypes.qry
+      })
+      .then(res => {
+        if (res.list.length > 0) {
+          const postTypes = res.list.map((record, index) => ({
```
NoraBitarSM, line 105: you are looping twice on the post types list
once here and once after getting the integration accounts list in getIntegrations( ).
You can loop only once for a better performance.

src/components/Shared/GlobalIntegrationGrid/index.js
NoraBitarSM, line 1: not sure if this code should be in GlobalIntegrationGrid/index.js
keep in mind that we will open this grid from a window on click of a button that we will add later on to the formshell actions array.

NoraBitarSM (changes requested):

src/components/Shared/GlobalIntegrationGrid/index.js
```diff
@@ -0,0 +1,178 @@
...
+  
+    const column = [
+      {
+        component: 'textfield',
+        label: _labels.ptName,
+        name: 'ptName',
+        props:{readOnly: true}
+      },
+      {
+        component: 'resourcecombobox',
```
NoraBitarSM, line 71: it should be a lookup not a combobox
sometimes there is over 1000 accounts. it's heavy to load them all in one step. lookups are better in this case
the lookup fills from GL.asmx/snapshotAC

src/components/Shared/GlobalIntegrationGrid/index.js
```diff
@@ -0,0 +1,178 @@
...
+        label: _labels.account,
+        name: 'accountRef',
+        props: {
+          endpointId: GeneralLedgerRepository.IntegrationAccounts.qry,
```
NoraBitarSM, line 75: endpoint to be changes

src/components/Shared/GlobalIntegrationGrid/index.js
NoraBitarSM, line 1: bottom buttons should be fixed on the screen's bottom right (without having to scroll to save)
the grid only should be scrollable.

src/components/Shared/GlobalIntegrationGrid/index.js
```diff
@@ -0,0 +1,178 @@
...
+    const column = [
+      {
+        component: 'textfield',
+        label: _labels.ptName,
+        name: 'ptName',
+        props:{readOnly: true}
+      },
+      {
+        component: 'resourcecombobox',
+        label: _labels.account,
```
NoraBitarSM, line 72: label is "account" not "accounts"

src/components/Shared/GlobalIntegrationGrid/index.js
NoraBitarSM, line 1: try to delete an account from a row and save, you will get this error
Error converting value {null} to type 'System.Int32'. Path 'integrationAccounts[6].accountId', line 1, position 762

NoraBitarSM (changes requested):

NoraBitarSM:
functionality tested

PR #206, opened 2024-04-25
currencies

omarZrayka23:
Each row currency should be filtered by the selected countryId from the monetary list (distinct currency)


NoraBitarSM:
you still have to work on the row's selection issue

NoraBitarSM:
show distinct currencies on monetary tab

PR #256, opened 2024-05-24
REM-295/ca-cash-inquiry

Alinadimmokdad:
ca-cash-inquiry 
Cash And Banks / Master Data/ Cash Inquiry

### DIFF
--- a/src/repositories/CashBankRepository.js
+++ b/src/repositories/CashBankRepository.js
@@ -54,6 +54,7 @@ export const CashBankRepository = {
     get: service + 'getABA'
   },
   CATransaction: {
-    qry: service + 'qryTRX'
+    page: service + 'qryTRX',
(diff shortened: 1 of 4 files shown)


src/repositories/CashBankRepository.js
```diff
@@ -52,5 +52,9 @@ export const CashBankRepository = {
   AccountBalance: {
     qry: service + 'qryABA',
     get: service + 'getABA'
+  },
+  CATransaction: {
+    page: service + 'qryTRX',
```
RebeccaFrem02, line 57: qry instead of page

RebeccaFrem02, line 57: check if you need the qry, if not please remove it

Alinadimmokdad: done

PR #267, opened 2024-05-31
Window error

omarZrayka23:
change height of  ErrorPage


src/providers/RequestsContext.js
```diff
@@ -39,12 +38,19 @@ function LoadingOverlay() {
...
+  const errorModel = useError()
   const [loading, setLoading] = useState(false)
 
   let isRefreshingToken = false
   let tokenRefreshQueue = []
 
+  function showError(props) {
+    if (errorModel) errorModel.stack(props)
+    else {
+      //show old error window
```
hadichahine, line 50: We should handle global errors here using <ErrorWindow>

src/providers/RequestsContext.js
hadichahine, line 1: Please fix ControlContext

hadichahine (changes requested):

hadichahine (approved):

PR #341, opened 2024-07-09
FixFilterCondition

Ghinwa-Arab:
fix isFilterMode condition

### DIFF
--- a/src/hooks/resource.js
+++ b/src/hooks/resource.js
@@ -11,7 +11,9 @@ export function useResourceQuery({ endpointId, filter, datasetId, queryFn, searc
 
   const isFilterMode =
     Object.keys(filters).length > 0 &&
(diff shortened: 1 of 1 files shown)


src/hooks/resource.js
```diff
@@ -9,7 +9,9 @@ export function useResourceQuery({ endpointId, filter, datasetId, queryFn, searc
 
   const isSearchMode = !!searchValue
 
-  const isFilterMode = Object.keys(filters).length > 0 && Object.values(filters).every(value => value)
+  const isFilterMode =
+    Object.keys(filters).length > 0 &&
+    Object.values(filters).every(value => value !== null && value !== undefined && value !== '')
```
hadichahine, line 14: replace `value !== ''` with `typeof value === 'String' && value.trim() !== ''`

hadichahine (changes requested):

PR #364, opened 2024-07-29
New sms request log

omarZrayka23:
i add  wrapText option, autoHeight for column and refactoring page


src/pages/sms-request-logs/index.js
```diff
@@ -0,0 +1,163 @@
...
+
+  const { formik } = useForm({
+    initialValues: {
+      moduleId: '10',
+      resourceId: ''
+    },
+    enableReinitialize: true,
+    validateOnChange: true,
+    validationSchema: yup.object({
+      moduleId: yup.string().required(' '),
```
RebeccaFrem02, line 33: we can remove the empty string here

src/pages/sms-request-logs/index.js
```diff
@@ -0,0 +1,163 @@
...
+    <VertLayout>
+      <Fixed>
+        <GridToolbar
+          maxAccess={access}
+          onSearch={value => {
+            filterBy('qry', value)
+          }}
+          labels={labels}
+        >
+          <Box sx={{ display: 'flex', width: '700px', justifyContent: 'flex-start', pt: 2, pl: 2, pb: 2 }}>
```
RebeccaFrem02, line 80: can't we use Grid instead of box so the styling would be less complicated?

PR #384, opened 2024-08-01
REM 350

imtithalhalalo:


src/components/Shared/Popper/PopperComponent.js
```diff
@@ -83,7 +83,8 @@ const PopperComponent = ({ children, anchorEl, open }) => {
       ref={popperRef}
       sx={{
         zIndex: '2 !important',
-        display: open && isVisible ? 'block' : 'none'
+        display: open && isVisible ? 'block' : 'none',
+        zIndex: 1300
```
hadichahine, line 87: acceptable for now but we should overhaul our stacking architecture

src/windows/index.js
```diff
@@ -6,6 +7,22 @@ import { v4 as uuidv4 } from 'uuid'
 const WindowContext = React.createContext(null)
 const ClearContext = React.createContext(null)
 
+function Overlay() {
```
hadichahine, line 10: There is an Overlay already inside Window. The task is to investigate why the original one doesn't block input.

hadichahine (changes requested):

PR #386, opened 2024-08-01
REM-424

RebeccaFrem02:
error when loading Argus for the first time


src/pages/_app.js
hadichahine, line 1: We should delete the Guard here. Ultimately having a GuestGuard on top of the app (below authcontext) and an AuthGuard in place of the current Guard.

hadichahine (changes requested):

PR #448, opened 2024-08-14
make disable submit +readOnly on status ! 1

Alinadimmokdad:
GL button in /journal-vouchers/ for testing


RebeccaFrem02:
image
add vertLayout and grow for the styling, and add a title for the window

PR #452, opened 2024-08-14
fix days : 0 instead of null

Alinadimmokdad:
fix days : 0 instead of null 
inn : 
/sa-payment-terms/


src/pages/sa-payment-terms/forms/PaymentTermsForms.js
```diff
@@ -44,6 +44,10 @@ export default function PaymentTermsForms({ labels, maxAccess, recordId }) {
       type: yup.string().required()
     }),
     onSubmit: async obj => {
+      if (!obj.days) {
+        obj.days = 0
```
imtithalhalalo, line 48: it cannot be in initial values as zero?

PR #460, opened 2024-08-16
check functionId before request

omarZrayka23:
click ADD, when i select the vendor group 0010, i get an error

in pu-vendors page

### DIFF
--- a/src/lib/fieldBehaviors.js
+++ b/src/lib/fieldBehaviors.js
@@ -25,7 +25,7 @@ const mergeWithMaxAccess = (maxAccess, field, fieldName) => {
   } else {
     if (field.mandatory || field.readOnly) {
       controls.push({
-        sgId: 18,
+        sgId: '',
         resourceId: '',
         controlId: fieldName,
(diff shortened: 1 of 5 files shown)


Ghinwa-Arab:
image

Reference not filling after save

Ghinwa-Arab:
image

this vendor group has no number range, reference in editMode should be editable (but take into consideration if vendor group is binded to number range reference should behave according to it even in editMode)

src/lib/fieldBehaviors.js
```diff
@@ -0,0 +1,77 @@
...
+  let controls = maxAccess.record.controls
+
+  let control = controls.find(obj => obj.controlId === fieldName)
+
+  if (control) {
+    control.accessLevel = field.mandatory ? MANDATORY : DISABLED
+  } else {
+    if (field.mandatory || field.readOnly) {
+      controls.push({
+        sgId: 18,
```
Ghinwa-Arab, line 28: why here we pass hard code sgId:18??

PR #461, opened 2024-08-16
Receipt vouchers

omarZrayka23:
1. when you select an account, sales person is filled from the spId of the account selection, ex: burger, account SVC00037 has a sales person
2.so the sales person is enabled only when the account reference is selected.
3.add save and clear button to the windowToolbar, the button is visible when we have the save button AND the clear button at the same time, the behaviour is like we clicked on save then we clicked on clear but here it is in one click only. button icon is attached .
4.info button is showing resourceId = 31301 but the correct one is 31304.


NoraBitarSM:
let's add a props to hide save and clear, for example Idon't need it in the report parameter browser 
image

NoraBitarSM:
please change the fields layout to become as in the screenshot 
image

PR #463, opened 2024-08-16
remove unused code

omarZrayka23:
1- Account id is empty

2- no need for info button here, delete it or make it enable

### DIFF
--- a/src/pages/company-info/index.js
+++ b/src/pages/company-info/index.js
@@ -80,7 +80,13 @@ const CompanyInfo = () => {
   }
 
   return (
-    <FormShell resourceId={ResourceIds.CompanyInfo} form={formik} infoVisible={false} maxAccess={maxAccess}>
+    <FormShell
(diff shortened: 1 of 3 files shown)


src/pages/company-info/index.js
```diff
@@ -89,8 +90,7 @@ const CompanyInfo = () => {
               value={formik.values?.accountId}
               onChange={formik.handleChange}
               readOnly={true}
-              onClear={() => formik.setFieldValue('posMsg', '')}
-              error={formik.errors && Boolean(formik.errors.posMsg)}
+              onClear={() => formik.setFieldValue('accountId', '')}
```
NoraBitarSM, line 93: the field is always readonly, why would you use onclear ?

NoraBitarSM (changes requested):

NoraBitarSM:
image
image

make the image attachment square, like the one we have in ASP. When I attach an image in react it's not fully visible

PR #487, opened 2024-08-21
Rem 510

omarZrayka23:
the current client relation button will open a readonly grid

we will add another button “add relation” next to the relation button in both client individual list and client corporate list

the ‘add relation’ will open a form with these fields:

client lookup (mandatory)

relation type combobox  (mandatory)

expiry date   (mandatory)

activation date  (mandatory) (by default today’s date)

otp verified (disabled checkbox)


src/components/Shared/ClientRelationForm.js
```diff
@@ -3,221 +3,198 @@ import FormShell from './FormShell'
...
-          rtId: '',
-          rtReference: '',
-          expiryDate: '',
-          activationDate: ''
-        }
-      ]
+      seqNo: 0,
+      parentId: '',
+      clientId: clientId,
+      parentName: '',
```
Alinadimmokdad, line 65: clientId: clientId || null ?

PR #496, opened 2024-08-21
Journal voucher

omarZrayka23:
i  removed the currencyId field , exRate and rateCalcMethod 

i add post button and i make field readonly onpost


Ghinwa-Arab:
image

save and clear btn should be disabled after posting

src/components/Shared/Buttons.js
```diff
@@ -233,15 +233,16 @@ export const getButtons = platformLabels => [
...
+    disabled: 'disabledSavedClear'
   },
   {
     key: 'Submit',
     label: platformLabels.Submit,
     condition: 'isSaved',
     onClick: 'onSave',
     color: '#4eb558',
-    disabled: 'disabledSubmit || isPosted || isClosed',
+    disabled: 'disabledSubmit || isPosted',
```
Ghinwa-Arab, line 245: why u remove ||closed? we have records are closed only and not posted so save should be disabled in this case

src/pages/journal-vouchers/forms/JournalVoucherForm.js
```diff
@@ -64,51 +62,74 @@ export default function JournalVoucherForm({ labels, access, recordId, general =
...
+          getData(response.recordId)
+        } else toast.success(platformLabels.Edited)
 
         invalidate()
       } catch (error) {}
     }
   })
+
   const editMode = !!formik.values.recordId || !!recordId
+  const isRaw = formik.values.status == 1
```
Ghinwa-Arab, line 75: `const isRaw = formik.values.status == 1
const isPosted = formik.values.status === 3`
u have 2 const where it's value is based on formik.values.status can we remove one of them to prevent duplicates?
for example readonly fields can be readOnly={editMode || isPosted}

check if it's possible and test it

src/components/Shared/Buttons.js
omarZrayka23: this changes is unintentional

PR #516, opened 2024-08-22
Rem 319

omarZrayka23:
in new mode click “save and clear” instead of “ save” this error appear:”Error converting value {null} to type 'ArgusDS.DocumentRelease.Constants.CharacteristicOperator'. Path 'characteristicOperator', line 1, position 58.


Ghinwa-Arab:
image

I tested the General Journal screen in 'New' mode. After filling the data, I clicked 'Save & Clear', but the form didn't clear after saving. It behaved as if it only performed the save operation.

omarZrayka23:
fixed in another branch

PR #521, opened 2024-08-23
REM-276

Alinadimmokdad:
add Integration button to fi-expense-types


Ghinwa-Arab:
image

add new record then save , info and integration btns still disable

PR #526, opened 2024-08-23
REM-523

imtithalhalalo:
'vehicles' page


Ghinwa-Arab:
image

Volume and weight accept only integers. If decimals present and less than 0.5, round it down to the nearest number, otherwise, round it up to the next number.

PR #528, opened 2024-08-23
REM-522

imtithalhalalo:
'drivers' page

### DIFF
--- a/src/pages/drivers/forms/DriversForm.js
+++ b/src/pages/drivers/forms/DriversForm.js
@@ -1,0 +1,126 @@
+import { Grid } from '@mui/material'
+import { useContext, useEffect } from 'react'
+import * as yup from 'yup'
+import FormShell from 'src/components/Shared/FormShell'
+import toast from 'react-hot-toast'
+import { RequestsContext } from 'src/providers/RequestsContext'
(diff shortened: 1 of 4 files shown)


src/pages/drivers/forms/DriversForm.js
```diff
@@ -0,0 +1,126 @@
...
+                value={formik.values.name}
+                required
+                maxAccess={maxAccess}
+                onChange={formik.handleChange}
+                onClear={() => formik.setFieldValue('name', '')}
+                error={formik.touched.name && Boolean(formik.errors.name)}
+              />
+            </Grid>
+            <Grid item xs={12}>
+              <CustomTextField
```
RebeccaFrem02, line 91: isn't it better if we use numberField for the cellphone number?

RebeccaFrem02:
image
the name of the page is not showing

src/pages/drivers/forms/DriversForm.js
```diff
@@ -0,0 +1,127 @@
...
+                name='cellPhone'
+                label={labels.cellPhone}
+                value={formik.values.cellPhone}
+                required
+                type='numeric'
```
omarZrayka23, line 96: remove it and add  phone={true}

PR #539, opened 2024-08-28
REM-452

imtithalhalalo:
stock count controller
page: 'sc-controller'


src/pages/sc-controller/forms/StockCountControllerForm.js
```diff
@@ -0,0 +1,114 @@
...
+      recordId: null,
+      plantId: null,
+      name: ''
+    },
+    maxAccess,
+    enableReinitialize: true,
+    validateOnChange: true,
+    validationSchema: yup.object({
+      name: yup.string().required(),
+      plantId: yup.string().required()
```
RebeccaFrem02, line 37: is plantId required because the plant combobox doesn't turn red and it doesn't have the "*"

image

Alinadimmokdad (approved):

PR #555, opened 2024-09-04
REM-454

imtithalhalalo:
new page: “sc-end-site-count”


FatimaSalami:
there is a problem in behaviour, when you make certain selections and post..
if you select them again the icon should be locked not as if it is first time to select them

src/pages/sc-end-site-count/forms/EndSiteCountForm.js
```diff
@@ -0,0 +1,139 @@
...
+    {
+      key: 'Lock',
+      condition: formik.values.statusName === 'Processed',
+      onClick: () => formik.handleSubmit(),
+      disabled: false
+    },
+    {
+      key: 'Unlock',
+      condition: formik.values.statusName !== 'Processed',
+      onClick: () => formik.handleSubmit(),
```
Alinadimmokdad, line 69: is it better to use status =1 or 2 ?

src/pages/sc-end-site-count/forms/EndSiteCountForm.js
```diff
@@ -0,0 +1,139 @@
...
+                maxAccess={access}
+              />
+            </Grid>
+            <Grid item xs={12}>
+              <CustomTextField
+                name='notes'
+                type='text'
+                label={_labels.notes}
+                readOnly
+                value={formik.values.notes}
```
Alinadimmokdad, line 127: is it readOnly ? 
if no use customTextArea

Alinadimmokdad (approved):

src/pages/sc-end-site-count/forms/EndSiteCountForm.js
imtithalhalalo: yes readOnly

src/pages/sc-end-site-count/forms/EndSiteCountForm.js
imtithalhalalo: same

NoraBitarSM: @imtithalhalalo don't compare based on the statusName, what if we're using argus in arabic ? @Alinadimmokdad is right

PR #559, opened 2024-09-10
remove clear from address form shell used as second tab

FatimaSalami:
we need to remove clear icon from address second tab
so i added the prop to be able to determine if we need the button or no..
sometimes we use address formshell for separate window so in this case we may need clear to maintain..

### DIFF
--- a/src/components/Shared/AddressForm.js
+++ b/src/components/Shared/AddressForm.js
@@ -5,7 +5,7 @@ import { RequestsContext } from 'src/providers/RequestsContext'
 import toast from 'react-hot-toast'
 import { ControlContext } from 'src/providers/ControlContext'
 
-const AddressForm = ({ recordId, address, setAddress = () => {}, editMode, onSubmit, isCleared = true }) => {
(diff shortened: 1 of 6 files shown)


src/components/Shared/AddressForm.js
```diff
@@ -5,7 +5,7 @@ import { RequestsContext } from 'src/providers/RequestsContext'
 import toast from 'react-hot-toast'
 import { ControlContext } from 'src/providers/ControlContext'
 
-const AddressForm = ({ recordId, address, setAddress = () => {}, editMode, onSubmit }) => {
+const AddressForm = ({ recordId, address, setAddress = () => {}, editMode, onSubmit, isCleared = true }) => {
```
hadichahine, line 8: Call it clearable. Let whatever is passed to plant window get passed to FormShell.

hadichahine (changes requested):

NoraBitarSM:
please remove the buttons from here also since u passed a props

image
image

hadichahine (approved):

PR #567, opened 2024-09-13
REM-453

imtithalhalalo:
page: 'sc-dtd'
required documentType
no need the useDocumentType or changeDT


FatimaSalami:
- if we click clear info button should be disabled
- if we click save and clear button should be disabled
- if we open edit mode info should be enabled

FatimaSalami (approved):

PR #586, opened 2024-09-18
use AG Grid Community only and cutomize copy select cell

omarZrayka23:
use AG Grid Community and add customise copy select  cell also, i can copied by doubleclick


hadichahine (approved):

NoraBitarSM:
where are the edit and delete icons ?
image

PR #597, opened 2024-09-23
Rem 547

omarZrayka23:
DATA GRID CHANGES TO AG GRID


NoraBitarSM:
1. when searching in the lookup and click TAB it should have stayed in the lookup field (datagrid level)
2. when searching in the lookup and selecting a row from the keyboard, when clicking TAB it should have selected the value focused on, instead of selecting the first one (lookup level)

PR #628, opened 2024-10-07
REM-607

imtithalhalalo:
page iv-barcodes

page items/items/barcodes tab this part from another task but it is add an add button above the grid and edit button to the rows
the form should be the same one used in barcodes screen except that the itemId, SKU and itemName are filled and the field is disabled when opened from here

the barcode field’s behavior will change

the barcodes will be filled automatically or no based on these conditions:

if the category doesn’t have a number range or the number range is external → barcode mandatory

if the category has number range and is not external → barcode optional and still enabled

if we save the barcode record without a value in the barcode field, the system will fill it automatically from the category’s number range ----> this one cannot be tested currently

if we save it with a value, the system will accept the value saved


src/components/Shared/GridToolbar.js
```diff
@@ -109,6 +110,11 @@ const GridToolbar = ({
                 />
               </Grid>
             )}
+            {middleSection && (
+              <Grid item sx={{ display: 'flex', justifyContent: 'flex-start', ml: 2 }}>
```
RebeccaFrem02, line 114: if we can remove the ml from here and add it to the item in the middle section

src/pages/iv-barcodes/Forms/BarcodesForm.js
```diff
@@ -0,0 +1,189 @@
...
+    ;(async function () {
+      try {
+        if (barcode) {
+          const res = await getRequest({
+            extension: InventoryRepository.Barcodes.get,
+            parameters: `_barcode=${barcode}`
+          })
+          formik.setValues(res.record)
+        }
+      } catch (e) {}
```
RebeccaFrem02, line 74: remove try catch

RebeccaFrem02:
image
you can fix the styling here everything will be same size and aligned

RebeccaFrem02:
image
the label of the checkbox is not showing

RebeccaFrem02:
image

RebeccaFrem02:
image
i changes the default qty of this record and saved, it didn't change in the outer grid

RebeccaFrem02:
image
i searched than chose a value in the lookup than applied, it didn't fetch properly

src/lib/barcodeFieldBehaviours.js
```diff
@@ -0,0 +1,98 @@
...
+
+const getData = async (getRequest, id) => {
+  try {
+    const res = await getRequest({
+      extension: SystemRepository.NumberRange.get,
+      parameters: `_recordId=${id}`
+    })
+
+    return res.record
+  } catch (error) {
```
RebeccaFrem02, line 12: remove try catch

PR #629, opened 2024-10-07
nowrap  text

omarZrayka23:
text overflow nowrap

after fixed  text is cut off

can u test on iv-items page

Screenshot 2024-10-08 at 12 52 01 PM

### DIFF
--- a/src/components/Shared/Table.js
+++ b/src/components/Shared/Table.js
@@ -452,7 +452,7 @@ const Table = ({
         setTimeout(() => setTooltipOpen(false), 500)
       })
     }
-    console.log(params)
+
     return (
       <Box>
         {tooltipOpen && (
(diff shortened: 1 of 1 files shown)


src/components/Shared/Table.js
```diff
@@ -485,7 +485,12 @@ const Table = ({
             },
             '&:focus': {
               outline: 'none'
-            }
+            },
+            ...(!params.colDef?.wrapText && {
```
imtithalhalalo, line 489: why do you need to check for wrapText? cannot be added directly?

omarZrayka23: wraptext is feature in ag grid and it effects on this fixed

src/components/Shared/Table.js
```diff
@@ -509,6 +514,7 @@ const Table = ({
       : []),
     ...filteredColumns.map(column => ({
       ...column,
+      flex: 1,
```
imtithalhalalo, line 517: if this by default then we no longer need to add it for each column now right?

FatimaSalami (approved):

PR #637, opened 2024-10-08
i fixed padding of cell in table and overlay label in cutsomtextfield

omarZrayka23:
i fixed padding of cell in table and overlay label in cutsomtextfield

u can do test iin v-items page


Ghinwa-Arab (commented):
no requested changes

PR #638, opened 2024-10-09
REM-620

imtithalhalalo:
prevent from adding the same reference twice in the grid

for the same type or for different

### DIFF
--- a/src/components/Shared/LOShipmentForm.js
+++ b/src/components/Shared/LOShipmentForm.js
@@ -341,7 +341,6 @@ export const LOShipmentForm = ({ recordId, functionId, editMode, totalBaseAmount
                           { seqNo: '1', id: 1, reference: '' }
                         ]
                       }
-                      onCellKeyDown={handleReferenceChange}
                       maxAccess={maxAccess}
                       allowAddNewLine={
                         !editMode &&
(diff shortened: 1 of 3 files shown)


RebeccaFrem02 (approved):

omarZrayka23:
When I click "tab," "onBlur" does not function. Change "onblur" to "onCellEditStop" instead.

src/components/Shared/LOShipmentForm.js
```diff
@@ -336,7 +364,8 @@ export const LOShipmentForm = ({ recordId, functionId, editMode, totalBaseAmount
                             maxLength: 20,
                             mandatory: true,
                             readOnly: editMode
-                          }
+                          },
+                          onBlur: (e, id) => handleReferenceChange(e, id)
```
omarZrayka23, line 368: do u need onBlur here?

src/components/Shared/LOShipmentForm.js
```diff
@@ -314,6 +341,7 @@ export const LOShipmentForm = ({ recordId, functionId, editMode, totalBaseAmount
                           { seqNo: '1', id: 1, reference: '' }
                         ]
                       }
+                      onCellKeyDown={handleReferenceChange}
```
omarZrayka23, line 344: Why is "handleReferenceChange" used in every cell?

omarZrayka23 (approved):

PR #667, opened 2024-10-15
Fix search

Alinadimmokdad:
Fix search in iv items index


Ghinwa-Arab:
image

First search for anything, then open the RPB window, and input 'IG-00028' in the SKU filter. After saving and applying, clear the search field. The grid resets to its original data, ignoring any selected parameters. **(It should return filtered data)**

Ghinwa-Arab (approved):

PR #677, opened 2024-10-17
REM-392

imtithalhalalo:
adding field set for checkboxs and fields
page: 'sales-settings'
add: sales defaults values and sales columns disable

### DIFF
--- a/src/pages/sales-settings/Form/SalesSettingsForm.js
+++ b/src/pages/sales-settings/Form/SalesSettingsForm.js
@@ -1,4 +1,4 @@
-import { Grid, Typography } from '@mui/material'
+import { Grid } from '@mui/material'
 import * as yup from 'yup'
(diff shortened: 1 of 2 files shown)


src/pages/sales-settings/Form/SalesSettingsForm.js
```diff
@@ -116,141 +116,149 @@ export default function SalesSettingsForm({ _labels, access }) {
...
+                  label={_labels.priceLevel}
+                  valueField='recordId'
+                  displayField={['reference', 'name']}
+                  displayFieldWidth={1}
+                  columnsInDropDown={[
+                    { key: 'reference', value: 'Reference' },
+                    { key: 'name', value: 'Name' }
+                  ]}
+                  values={formik?.values}
+                  maxAccess={access}
```
Alinadimmokdad, line 134: why are we using ? before formik values

PR #682, opened 2024-10-18
REM-628

imtithalhalalo:
page: 'ct-exchange-rates'
we cannot save with min > rate > max

it should only allow min <= rate <= max


RebeccaFrem02:
image
it is not saving even if the min<rate<max

PR #689, opened 2024-10-22
REM-642

RebeccaFrem02:
Login page:

show a loading when clicking on the login button 

when we sign out it seems that the login page is loading twice

if an error is returned when loading the login page, show it (ex: error in /KVS.asmx/qryLBL2)


src/providers/AuthContext.js
```diff
@@ -35,45 +35,47 @@ const AuthProvider = ({ children }) => {
...
+    const accountName = matchHostname ? matchHostname[1] : 'byc-deploy'
 
-        setCompanyName(response.data.record.companyName)
-        setGetAC(response)
-        window.localStorage.setItem('apiUrl', response.data.record.api)
-      } catch (error) {
-        console.error('Error fetching data:', error)
-      }
+    try {
+      const response = await axios.get(`${process.env.NEXT_PUBLIC_AuthURL}/MA.asmx/getAC?_accountName=${accountName}`)
```
Ghinwa-Arab, line 60: can we remove "MA.asmx/getAC" hard code and import it from repository?

Ghinwa-Arab (approved):

PR #701, opened 2024-10-24
REM-655

RebeccaFrem02:
add new readonly field that shows the generated reference based on the correspondent’s number range when posting the outwards transfer


src/pages/outwards-transfer/Tabs/OutwardsTransferForm.js
```diff
@@ -181,406 +181,432 @@ export default function OutwardsTransferForm({ labels, maxAccess, recordId }) {
...
-                    name='clientId'
-                    label={labels.Client}
-                    form={formik}
-                    required
+                <Grid item xs={3}>
+                  <CustomTextField
+                    name='soaRef'
+                    label={labels.soaRef}
+                    value={formik?.values?.soaRef}
+                    maxAccess={false && maxAccess}
```
imtithalhalalo, line 253: why do you need false?

imtithalhalalo (approved):

PR #714, opened 2024-10-28
REM-614

Alinadimmokdad:
fix invalidate for physical form


omarZrayka23:
When I modified the weight value and submitted it, the value in the index table did not change, and it was not requested. also it did not request getRequest onmouseLeave

omarZrayka23:
density did not request getRequest onmouseLeave

omarZrayka23 (approved):

PR #718, opened 2024-10-30
Quick search

imtithalhalalo:
adding quick search for production queue and items in production pages


Ghinwa-Arab:
image

After performing a search and clearing it, the grid data does not refresh automatically, even when pressing enter or clicking the refresh button

Ghinwa-Arab:
image

I'm searching by sku and itemName, both not working at all

Ghinwa-Arab:
image
image

see initially we have 19 records, when I search it returns 100 records! how?

PR #723, opened 2024-11-01
REM-535

imtithalhalalo:
adding reopen the record
 Post button should be disabled until we close the record 
 try to add a new record, add a line then delete it, the grid become disabled and it shouldn't
 try to print the template this error appear: Unexpected character encountered while parsing value: <. Path '', line 0, position 0

### DIFF
--- a/src/pages/outbound-transportation/forms/OutboundTranspForm.js
+++ b/src/pages/outbound-transportation/forms/OutboundTranspForm.js
@@ -278,7 +278,7 @@ export default function OutboundTranspForm({ labels, maxAccess: access, recordId
       key: 'Reopen',
       condition: isClosed,
       onClick: onReopen,
-      disabled: !isClosed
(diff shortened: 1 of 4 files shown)


src/pages/outbound-transportation/forms/OutboundTranspForm.js
```diff
@@ -75,66 +75,64 @@ export default function OutboundTranspForm({ labels, maxAccess: access, recordId
...
+          hasDuplicates = true
+          break
         }
+        soIdSet.add(order.soId)
+      }
 
-        if (hasDuplicates) {
-          //console.log('Duplicate soId found. Operation aborted.')
+      if (hasDuplicates) {
+        //console.log('Duplicate soId found. Operation aborted.')
```
Alinadimmokdad, line 111: you need this comment ?

PR #727, opened 2024-11-01
REM-680

RebeccaFrem02:
we need to introduce the 2nd FA that is an OTP form loaded if the user’s isEFAEnabled is true
SY.US.isEFAEnabled


src/components/Shared/OTPAuthentication.js
```diff
@@ -0,0 +1,151 @@
...
+import toast from 'react-hot-toast'
+import axios from 'axios'
+import { useAuth } from 'src/hooks/useAuth'
+import { useError } from 'src/error'
+import { ControlContext } from 'src/providers/ControlContext'
+import { AccessControlRepository } from 'src/repositories/AccessControlRepository'
+
+const OTPAuthentication = ({ loggedUser, onClose, window }) => {
+  const [otp, setOtp] = useState(['', '', '', '', '', ''])
+  const [error, setError] = useState('')
```
Alinadimmokdad, line 13: this could be useState([])

RebeccaFrem02: error is  string it is better if the initial value is ' '

src/components/Shared/OTPAuthentication.js
```diff
@@ -0,0 +1,151 @@
...
+import styles from '../../../styles/phoneVerification.module.css'
+import toast from 'react-hot-toast'
+import axios from 'axios'
+import { useAuth } from 'src/hooks/useAuth'
+import { useError } from 'src/error'
+import { ControlContext } from 'src/providers/ControlContext'
+import { AccessControlRepository } from 'src/repositories/AccessControlRepository'
+
+const OTPAuthentication = ({ loggedUser, onClose, window }) => {
+  const [otp, setOtp] = useState(['', '', '', '', '', ''])
```
Alinadimmokdad, line 12: this could be useState([]) ?

src/components/Shared/OTPAuthentication.js
```diff
@@ -0,0 +1,151 @@
...
+        })
+        .catch(error => {
+          showError({
+            message: error,
+            height: error.response?.status === 404 || error.response?.status === 500 ? 400 : ''
+          })
+          if (throwError) reject(error)
+        })
+    } else {
+      setError('All Fields Required')
```
FatimaSalami, line 57: isn't it better to get the error msgs from labels?

FatimaSalami (approved):

PR #739, opened 2024-11-05
REM-650

RebeccaFrem02:
the otp field should not be mandatory when saved with empty value


src/pages/ct-default/form/CtDefaults.js
```diff
@@ -39,7 +39,7 @@ const CtDefaults = ({ _labels, access }) => {
     enableReinitialize: true,
     validateOnChange: true,
     validationSchema: yup.object({
-      'otp-expiry-time': yup.number().min(30).max(120)
+      'otp-expiry-time': yup.number().min(30).max(120).nullable(true)
     }),
```
Alinadimmokdad, line 43: there is an error in the functionality , try to add a number then blur then press the clear ( X ) button it will not remove it

PR #778, opened 2024-11-14
Cash invoice

omarZrayka23:
i created a payment component that is shared and contains columns and a data grid.The Formik file contains the initial value and validation paymentGrid

cash type is not duplicate in rows 
in rt-receipt-voucher , total amount in grid  = total amount in field above(Amount)
in cash-invoice form  , total amount in  amount grid  = total lcAmount in operations grid

Only the cash type row is included in the returnAmount value; the other return amount is zero.

When a value is entered in the id number field on a cash invoice, onBlur automatically fills in other fields with the first number,  lastName id types, ...  .For example, 1999999999 or 1112223334 are autofill ID types only.


Hassan-SoftMachine (commented):
click fetch => type confirm id => click save => got this error => 
Screenshot 2024-12-10 at 9 26 24 AM

fill form values => click saved => confirm id => got this error => 
Screenshot 2024-12-10 at 9 27 48 AM

PR #784, opened 2024-11-18
lost options of lookup when i click arrowDown by keyboard

omarZrayka23:
lost options of lookup when i click arrowDown by keyboard
When using the keyboard's fast select option in the data grid, the value remains unchanged.
page : sales order


RebeccaFrem02:
image
image


the value was removed when the lookup lost focus

omarZrayka23:
> he value was removed when the lookup lost focus,
because The value of owoRef is not set in onchange.

PR #787, opened 2024-11-19
masterSource

imtithalhalalo:
Moving master source from FormShell props to be sent to the action button Integration Action
please review the pages that has change in master source


omarZrayka23:
Screenshot 2024-11-27 at 3 05 51 PM


The address in the textArea field has to be verified in the window

omarZrayka23:
Screenshot 2024-11-27 at 3 21 41 PM

src/pages/fi-accounts/forms/AccountsForm.js
RebeccaFrem02, line 1: image
the clear button of the lookup is not working

RebeccaFrem02: check the onClear function

RebeccaFrem02: image
check the functionality of the clear function of this field

omarZrayka23 (approved):

PR #794, opened 2024-11-21
Rm 102

Alinadimmokdad:
default page without get Request : 
gl-sync-org-bal
just post request

### DIFF
--- a/src/pages/gl-sync-org-bal/form/GlSyncForm.js
+++ b/src/pages/gl-sync-org-bal/form/GlSyncForm.js
@@ -1,0 +1,110 @@
+import { Grid } from '@mui/material'
+import * as yup from 'yup'
+import { useContext } from 'react'
+import FormShell from 'src/components/Shared/FormShell'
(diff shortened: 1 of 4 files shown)


src/pages/gl-sync-org-bal/form/GlSyncForm.js
```diff
@@ -0,0 +1,110 @@
...
+                columnsInDropDown={[
+                  { key: 'reference', value: 'Reference' },
+                  { key: 'name', value: 'Name' },
+                  { key: 'flName', value: 'FL Name' }
+                ]}
+                values={formik.values}
+                maxAccess={access}
+                onChange={(event, newValue) => {
+                  formik.setFieldValue('currencyId', newValue?.recordId || null)
+                }}
```
RebeccaFrem02, line 90: even if the field is not mandatory i think we still ned to hve the error prop

RebeccaFrem02 (approved):

PR #797, opened 2024-11-26
RM-104

Alinadimmokdad:
gl-fy-opening deafult page without get request , just post request


omarZrayka23:
The color does not turn red when I clear the combobox and click save.

omarZrayka23:
do u need clear button??

omarZrayka23:
decrease height of window

omarZrayka23 (approved):

PR #803, opened 2024-11-27
RM-110

Alinadimmokdad:
add new combo Outer Grid Days to system-defaults + refactor some code


src/pages/lo-defaults/form/LoDefault.js
```diff
@@ -62,54 +63,55 @@ const LoDefault = ({ _labels, access }) => {
...
+                columnsInDropDown={[
+                  { key: 'reference', value: 'Reference' },
+                  { key: 'name', value: 'Name' }
+                ]}
+                values={formik.values}
+                valueField='recordId'
+                displayField='name'
+                maxAccess={access}
+                onChange={(event, newValue) => {
+                  if (newValue) {
```
imtithalhalalo, line 92: remove if else and added it as ternary operation

PR #808, opened 2024-11-28
RM-61

imtithalhalalo:
receipt vouchers bugs + refactoring
1- add a long number in amount field and save this error appear:”Arithmetic overflow error converting float to data type numeric. The statement has been terminated. 

2- fix post button design AND I need unpost in the screen

3- in new mode save and clear is not working, is should click 2 times, first time it's save the second time the data is cleared


Alinadimmokdad:
image
On save and Clear

Alinadimmokdad:
image
after save and clear the account ref is not cleared

Alinadimmokdad:
image
the error still while adding data

Alinadimmokdad (approved):

PR #817, opened 2024-12-02
Rm 127

Alinadimmokdad:
default page get & post data : 
gl-system-params


src/pages/gl-system-params/form/SystemParamsForm.js
```diff
@@ -0,0 +1,370 @@
...
+                name='GLDOEGainAccountId'
+                label={_labels.glDoeGain}
+                form={formik}
+                required
+                displayFieldWidth={2}
+                valueShow='GLDOEGainAccountref'
+                secondValueShow='GLDOEGainAccountname'
+                maxAccess={access}
+                onChange={(event, newValue) => {
+                  formik.setFieldValue('GLDOEGainAccountId', newValue ? newValue.recordId : '')
```
Hassan-SoftMachine, line 168: this structure is repetitive, there are multiple in this single file, you can create a custom onChange something like 
` const handleChange = fieldArray => {
    fieldArray.forEach(({ name, value }) => {
      formik.setFieldValue(name, value)
    })
  }`

and inside the onchange it could be something like 
`const { recordId = '', accountRef = '', name = '' } = newValue || {};
handleChange([{name:'GLDOEGainAccountId',value:recordId},{name:'GLDOEGainAccountref',value:accountRef},{name:'GLDOEGainAccountname',value:name}])
`

it would be more readable and more maintainable and not very repetitive

src/pages/gl-system-params/form/SystemParamsForm.js
```diff
@@ -0,0 +1,370 @@
...
+                displayField='name'
+                name='GLDOELossAccountId'
+                label={_labels.glDoeLoss}
+                form={formik}
+                required
+                displayFieldWidth={2}
+                valueShow='GLDOELossAccountref'
+                secondValueShow='GLDOELossAccountname'
+                maxAccess={access}
+                onChange={(event, newValue) => {
```
Hassan-SoftMachine, line 188: use the custom onChange here too, and every where you think its applicable

Hassan-SoftMachine (approved):

PR #829, opened 2024-12-05
 column width

omarZrayka23:
increase columns width so all the column title is readable


imtithalhalalo:
there is no changes related to column width only formatting can you check please?

imtithalhalalo (dismissed):

NoraBitarSM (changes requested):
when I imported a file the columns width decreased changed and they are not visible anymore
image

PR #832, opened 2024-12-06
Rem 712

imtithalhalalo:
changes for outwards return
adding new fields and changing layout


NoraBitarSM:
1. in the outwards lookup, dhow in the dropdown: reference, amount, commission, correspondentName, clientName, currencyRef.
2. outward reference is empty in the outer grid
3. don't calculate the amount fields set values unledd requested by is selected because the calculation depends on it 
image

4. when I edited or closed the record, lc amount and amount values disappeared 
image
image
5. the rate was 5.27 from the outwards transfer, why is it showing 5.25 ? 
image
6. in new mode set the settlement status by default always "Waiting For Correspondent Response" and readonly

PR #834, opened 2024-12-06
gocIssue

Ghinwa-Arab:
currency combo in both screens should show all currencies, and on selection request getCCU to get goc value


src/pages/credit-invoice/Forms/CreditInvoiceForm.js
```diff
@@ -740,6 +738,14 @@ export default function CreditInvoiceForm({ _labels, access, recordId, plantId,
       width: 130
     }
   ]
+  async function getCurrencyGoc(currencyId) {
+    const res = await getRequest({
+      extension: RemittanceSettingsRepository.CorrespondentCurrency.get,
+      parameters: `_corId=${formik.values.corId}&_currencyId=${currencyId}`
+    })
+
+    return res?.record?.goc
```
NoraBitarSM, line 747: return res?.record and then get the goc value in the update function. if the res.record is null then you can consider that goc is false

src/pages/credit-order/Forms/CreditOrderForm.js
```diff
@@ -598,13 +590,21 @@ export default function CreditOrderForm({ labels, access, recordId, plantId, use
...
     }
   ]
 
+  async function getCurrencyGoc(currencyId) {
+    const res = await getRequest({
+      extension: RemittanceSettingsRepository.CorrespondentCurrency.get,
+      parameters: `_corId=${formik.values.corId}&_currencyId=${currencyId}`
+    })
+
+    return res?.record?.goc
```
NoraBitarSM, line 605: return res?.record and then get the goc value in the update function. if the res.record is null then you can consider that goc is false

NoraBitarSM (changes requested):

NoraBitarSM (approved):

PR #840, opened 2024-12-10
scanner image

omarZrayka23:
page: client individual list
Pressing the button ''Scanner' causes the scanner to display an image, which is then saved when I hit "Submit."


NoraBitarSM:
please enable these 3 buttons only in new mode or in the edit client file page, same as the other fields 
image

Hassan-SoftMachine (approved):

PR #841, opened 2024-12-10
add cache of combobox in data grid

omarZrayka23:
datagrid / combobox caching

Combobox caching prevents the combobox from making a request to save data in each row.


Hassan-SoftMachine (approved):
please lmk how to test it before merging

Hassan-SoftMachine (commented):
when clicking on the combobox its filling by default 888 for some reason

PR #847, opened 2024-12-11
adding 2 checkbox rmitem and inactive

FatimaSalami:
in item screen 2 checkboxes were added.
old checkboxes on change was edited because this new way instead of handle change is better.
window size is increased because we don't need to see scroll

### DIFF
--- a/src/pages/iv-items/forms/ItemsForm.js
+++ b/src/pages/iv-items/forms/ItemsForm.js
@@ -67,7 +67,9 @@ export default function ItemsForm({ labels, maxAccess: access, setStore, store, 
       spfId: '',
       categoryName: '',
       defSaleMUId: '',
-      pgId: ''
+      pgId: '',
+      isInactive: false,
+      rmItem: false
     },
     maxAccess,
     enableReinitialize: true,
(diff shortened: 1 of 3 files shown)


src/pages/iv-items/forms/ItemsForm.js
```diff
@@ -469,7 +480,9 @@ export default function ItemsForm({ labels, maxAccess: access, setStore, store,
                       <Checkbox
                         name='ivtItem'
                         checked={formik.values.ivtItem}
-                        onChange={formik.handleChange}
+                        onChange={event => {
```
omarZrayka23, line 483: u can used formik.handleChange and in edit mode if value equal null u can use || false

Alinadimmokdad (approved):

src/pages/iv-items/window/ItemWindow.js
```diff
@@ -55,28 +55,28 @@ const ItemWindow = ({ recordId, labels, sku, itemName, msId, maxAccess, setTitle
           setFormikInitial={setFormikInitial}
         />
       </CustomTabPanel>
-      <CustomTabPanel height={660} index={1} value={activeTab}>
+      <CustomTabPanel height={670} index={1} value={activeTab}>
```
omarZrayka23, line 58: Increase the height using the Stack component's props on the index page.

omarZrayka23 (approved):

PR #855, opened 2024-12-16
RM-169

imtithalhalalo:
fixing plantId and siteId to be required
screen sc-cycle-counts


src/pages/sc-cycle-counts/Forms/CycleCountsForm.js
RebeccaFrem02, line 1: image
there is no need for the scroll here, please check the grids to remove the scrollbar

RebeccaFrem02 (approved):

PR #859, opened 2024-12-16
Rm 71

Alinadimmokdad:
md-asset-depreciation/
fix post unpost behavior


imtithalhalalo:
Screenshot 2024-12-17 081219
i cannot add any row

PR #863, opened 2024-12-17
REM-717

omarZrayka23:
add Checkbox in yakeen form
page : 'client-individual-list'
click fetch button after fill birthday and id number  ex:  id number: 1234567891


Alinadimmokdad:
image

PR #868, opened 2024-12-19
RM-175

Alinadimmokdad:
fix error in data grid columns /security-groups/


RebeccaFrem02:
image

i clicked on the clear button and saved, but when i reopened it i refound the data

PR #876, opened 2024-12-26
RM-55

imtithalhalalo:
fixing barcodes clear
barcode screen, edit any record click “save and clear” nothing happen, all data is not cleared


RebeccaFrem02:
image
barcode should be mandatory

Ghinwa-Arab:
image

Initially, both descriptions are set to itemName. However, during the save process, the API is rejecting sales description, possibly because it exceeds the maximum allowed length. Please consult with a backend team member for further clarification.

Ghinwa-Arab (approved):

PR #888, opened 2025-01-02
Rm 176

omarZrayka23:
page: physical-count-itemDetails
jump to nextLine, duplicate, maxAccess and defaultValue.
 
1-check video 
 
2- If disableLookup, the sku field is textField; otherwise, it is lookup;
    if jumpToNextLine was checked, when click tab on the sku add new row and jump to next line.
    if default qty is checked as true, qty takes the default value; 
   if duplicate is true, avoid duplicate sku in the data grid.

please when testing test for jump to next line, duplicate and def qty for both text field and lookup
test for combined scenarios
and focus on basic grid functionality
note: jump to next line neglect current line after data is field on sku selection and move to next line directly (not like normal tab)

and of course compare the behaviour to asp (if anything is unclear don't hesitate to ask)


Ghinwa-Arab:
image
The default quantity is being updated when a row is added, but this is incorrect. It should be set when an SKU is selected, not when a new row is added.

Ghinwa-Arab:
image

if sku not valid pls clear the row values

Ghinwa-Arab:
image
If you press 'Clear All' then no empty row is displayed (I cannot add new row)

imtithalhalalo:
Screenshot 2025-01-07 082951
site has empty data I cannot add to the data grid

imtithalhalalo:
Screenshot 2025-01-07 084146
when I cleared using clear all and then filled the fields I wasn't allowed to add to the data grid

imtithalhalalo:
Screenshot 2025-01-07 084338
first row can show sku not required?

omarZrayka23:
> Collaborator

fixed

Ghinwa-Arab:
image
image

default qty not checked and still on sku selection it displaying qty=1

Ghinwa-Arab:
In case qty is not checked and jump to next line checked then it displaying default qty on sku selection, otherwise if jump to next line not checked it's working normally

omarZrayka23:
> image In the system checks, check 'default quantity' and 'jump to the next line'. Then, fill the grid with any data, press "Clear Grid," and try to set the data again. When selecting an SKU, the default quantity updates in the new added row instead of the previous one.

fixed

omarZrayka23:
> image If you press 'Clear Grid' then no empty row is displayed (I cannot add new row)

fixed

Ghinwa-Arab (approved):

PR #914, opened 2025-01-10
AR-2657

imtithalhalalo:
fixing siteId
the default value of site is from SY.getDE and of key “de_siteId”
and the combo should be readonly


src/pages/iv-replenishment/forms/IvReplenishementsForm.js
```diff
@@ -85,37 +85,45 @@ const IvReplenishementsForm = ({ labels, maxAccess, setStore, store }) => {
...
   const editMode = !!recordId
 
+  async function getDefaultSiteId() {
+    if (editMode) {
+      return
+    }
+
+    const defaultSiteId = defaultsData?.list?.find(({ key }) => key === 'de_siteId')
+
+    if (defaultSiteId?.value && !formik.values.fromSiteId)
```
FatimaSalami, line 95: && !formik.values.fromSiteId is there a need for this condition?

FatimaSalami (approved):

PR #938, opened 2025-01-22
customNumber field

omarZrayka23:
page: clients-list
create customPhoneNumber


src/pages/clients-list/forms/ClientTemplateForm.js
```diff
@@ -966,7 +989,6 @@ const ClientTemplateForm = ({ recordId, labels, plantId, maxAccess, allowEdit =
                           required
                           onChange={e => {
                             setIsValidatePhoneClicked(false)
-                            formik.setFieldTouched('idNo', true)
```
FatimaSalami, line 969: why is this removed?

FatimaSalami:
when we clear using backspace + can't be deleted 
while when we clear using the button everything is deleted but the icon of the country is note changed(not deleted nor returned to default one) i don't know if this is the best behaviour

FatimaSalami:
image
when opened in edit mode 
i can still change the selection of country icon thus changing the code
it is not harmful because we can't save 
but is it accepted to still be able to change?

src/pages/clients-list/forms/ClientTemplateForm.js
```diff
@@ -1528,12 +1552,12 @@ const ClientTemplateForm = ({ recordId, labels, plantId, maxAccess, allowEdit =
                           maxLength='15'
```
FatimaSalami, line 1552: why is && true used in readOnly?

src/pages/clients-list/forms/ClientTemplateForm.js
omarZrayka23: it makes textField slowly

omarZrayka23:
> i don't know if this is the best behavi

yes, here have default country from system default

FatimaSalami:
Video sent inbox concerning the clear behaviour and the touch of id change

src/pages/clients-list/forms/ClientTemplateForm.js
FatimaSalami: it affects the behaviour when editing then deleting the written numbers (in video)

src/pages/clients-list/forms/ClientTemplateForm.js
FatimaSalami: above it also in the other phone number fix it please

FatimaSalami (approved):

PR #941, opened 2025-01-23
AR-2654

imtithalhalalo:
sa-clients we can save without acquisitionDate it was showing like that
Could not convert string to DateTime: /Date(null)/. Path 'acquisitionDate', line 1, position 422.


NoraBitarSM:
I am selecting "today" in the dropdown and after saving it is showing yesterday's date

PR #942, opened 2025-01-24
AR-2784

imtithalhalalo:
exRate error screen credit note
Error converting value (null, to type 'System.Decimal. Path 'exRate", line position 224.
currency only in ca-adjustment is not showing default value


NoraBitarSM:
by default or if there is no data returned from MC.getDRI then exRate and rateCalcMethod should be 1

Ghinwa-Arab (approved):

PR #943, opened 2025-01-24
AR-2776

imtithalhalalo:
payment voucher is not saving
react/account : anthonys
payment voucher screen, try to edit a raw activity
change the date and save
the submit button is not working and it should save


NoraBitarSM:
add the data normally
then change the date
and change the amount
the base amount is nor recalculated on the last amount change

image

PR #947, opened 2025-01-27
AR-2666

imtithalhalalo:
fixing cash account bug always showing as null
screen: fi-receipt-vouchers

### DIFF
--- a/src/pages/fi-receipt-vouchers/forms/ReceiptVoucherForm.js
+++ b/src/pages/fi-receipt-vouchers/forms/ReceiptVoucherForm.js
@@ -207,8 +207,10 @@ export default function ReceiptVoucherForm({ labels, maxAccess: access, recordId
     ;(async function () {
       if (formik.values.recordId) await getCashAccount()
       await getData(formik.values.recordId)
-      getUserDefaultsData()
-      getDefaultsData()
+      if (!editMode) {
+        getUserDefaultsData()
+        getDefaultsData()
+      }
     })()
   }, [])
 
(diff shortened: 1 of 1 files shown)


src/pages/fi-receipt-vouchers/forms/ReceiptVoucherForm.js
```diff
@@ -204,8 +204,8 @@ export default function ReceiptVoucherForm({ labels, maxAccess: access, recordId
   }
 
   useEffect(() => {
-    if (formik.values.recordId) getCashAccount()
     ;(async function () {
+      if (formik.values.recordId) await getCashAccount()
       await getData(formik.values.recordId)
       getUserDefaultsData()
```
FatimaSalami, line 210: this is wrong
default data should be filled only in new mode
so either we put a condition or we call them at the beginning

because now they will overwrite existing data
try to edit a record select different plant and save then close and open the record again, you will get the default plant that usually appears in new mode!

src/pages/fi-receipt-vouchers/forms/ReceiptVoucherForm.js
```diff
@@ -205,8 +206,10 @@ export default function ReceiptVoucherForm({ labels, maxAccess: access, recordId
 
   useEffect(() => {
     ;(async function () {
-      if (formik.values.recordId) await getCashAccount()
-      await getData(formik.values.recordId)
+      if (recordId) {
+        await getCashAccount()
+        await getData(recordId)
+      }
       if (!editMode) {
```
FatimaSalami, line 213: let it be else

FatimaSalami (approved):

PR #949, opened 2025-01-27
gl segment

imtithalhalalo:
gl segment changes
adding error when there is no segment chosen
fixing bug when adding a record it was removing everything and not just adding
changing the qry to use another key

### DIFF
--- a/src/pages/gl-segments/index.js
+++ b/src/pages/gl-segments/index.js
@@ -16,17 +16,19 @@ import { Grid } from '@mui/material'
 import { SystemRepository } from 'src/repositories/SystemRepository'
 import SegmentForm from './form/SegmentForm'
 import CustomComboBox from 'src/components/Inputs/CustomComboBox'
+import { useError } from 'src/error'
 
 const Segments = () => {
   const { getRequest, postRequest } = useContext(RequestsContext)
   const { platformLabels } = useContext(ControlContext)
   const [data, setData] = useState([])
   const [store, setStore] = useState([])
+  const { stack: stackError } = useError()
 
   const { stack } = useWindow()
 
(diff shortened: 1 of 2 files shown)


Ghinwa-Arab:
image
no selected segment and still opening

src/pages/gl-segments/index.js
```diff
@@ -16,17 +16,19 @@ import { Grid } from '@mui/material'
...
   const { platformLabels } = useContext(ControlContext)
   const [data, setData] = useState([])
   const [store, setStore] = useState([])
+  const { stack: stackError } = useError()
 
   const { stack } = useWindow()
 
   async function fetchGridData(options = {}) {
-    if (formik.values.segmentId) {
+    if (formik.values.segmentId !== null && formik.values.segmentId !== undefined) {
```
Ghinwa-Arab, line 31: it can be  if(formik.values.segmentId)
pls try

src/pages/gl-segments/index.js
```diff
@@ -96,19 +96,23 @@ const Segments = () => {
...
-        obj,
-        maxAccess: access,
-        formikSegmentId: formik.values.segmentId,
-        fetchGridData
-      },
-      width: 500,
-      height: 260,
-      title: _labels.segment
-    })
+    if (formik.values.segmentId === null || formik.values.segmentId === undefined) {
```
Ghinwa-Arab, line 99: it can be if (!formik.values.segmentId)
pls try

src/pages/gl-segments/index.js
imtithalhalalo: no because there is segmentId is zero

PR #954, opened 2025-01-28
RM-155

omarZrayka23:
in combobox in grid when we select my the mouse we lose focus, it should be left in the grid current cell (can be tested in Delivery Order)

in combobox when we search in the combo and still have 1 list item appearing in drop down for example, if we press tab this item is not selected, instead the first item in initial list without filtering is being filled.

summary: fixed save focus after mouse and auto select  first option  when do onBlur after search.

 please check all behaviors of combobox


imtithalhalalo:
Screenshot 2025-01-30 161007
when I clicked site i lost focus after third time clicking

imtithalhalalo:
Screenshot 2025-01-30 161234
when I tried to select site third time I lost the focus on the record

imtithalhalalo:
Screenshot 2025-01-30 161415
also here when I searched for byblos I am still getting first record Beirut

FatimaSalami (approved):

imtithalhalalo (approved):

PR #968, opened 2025-01-30
Ar 2792

FatimaSalami:
add default shipment and bill buttons
you can compare the behaviour to asp
these 2 buttons are added in address form 
they are disabled when selected 
you can add another address and set the defaults from it so they are enabled in the other records 

setstore is used so that store is updated to establish this performance of disabling and enabling buttons (depends on comparing address id with ship or bill address ids)
get request is used because we need to send the whole client on save with the correct ship to and bill to address
the state is used to disable the button directly


src/pages/sa-clients/forms/AddressTab.js
```diff
@@ -54,7 +54,8 @@ const AddressTab = ({ store, maxAccess, labels, editMode, ...props }) => {
         maxAccess,
         editMode,
         recordId: id,
-        clientId: recordId,
+        store,
+        setStore,
         getAddressGridData: getAddressGridData
```
omarZrayka23, line 59: you can use direct getAddressGridData

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -30,10 +32,70 @@ const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, pr
         window.close()
       })
     }
-    getAddressGridData(clientId)
+    getAddressGridData(store.recordId)
```
omarZrayka23, line 35: why use await and .then , u can use .then only

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -5,15 +5,17 @@ import { ControlContext } from 'src/providers/ControlContext'
 import AddressForm from 'src/components/Shared/AddressForm'
 import { SaleRepository } from 'src/repositories/SaleRepository'
 
-const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, props }) => {
+const ClientsAddressForm = ({ getAddressGridData, recordId, window, props, store, setStore }) => {
   const [address, setAddress] = useState()
-  const { postRequest } = useContext(RequestsContext)
+  const { postRequest, getRequest } = useContext(RequestsContext)
   const { platformLabels } = useContext(ControlContext)
+  const [isDefaultShip, setIsDefaultShip] = useState(false)
```
omarZrayka23, line 12: do u need state , u can check shipAddressId in store

omarZrayka23, line 12: also u can check billAddressId from store  and clear state

FatimaSalami: in react when we update the store instantly it is not changes, so it will still be null

FatimaSalami: in react when we update the store instantly it is not changes, so it will still be null

src/pages/sa-clients/forms/AddressTab.js
```diff
@@ -54,7 +54,8 @@ const AddressTab = ({ store, maxAccess, labels, editMode, ...props }) => {
         maxAccess,
         editMode,
         recordId: id,
```
RebeccaFrem02, line 56: name it directly recordId from the prop and send it only recordId

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -5,15 +5,17 @@ import { ControlContext } from 'src/providers/ControlContext'
...
+  const [isDefaultBill, setIsDefaultBill] = useState(false)
```
RebeccaFrem02, line 13: you can replace both these states with dynamic variables

check how we change the isPosted an isClosed in other pages

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -5,15 +5,17 @@ import { ControlContext } from 'src/providers/ControlContext'
...
 
   const onSubmit = async obj => {
     if (obj) {
       const data = {
-        clientId: clientId,
+        clientId: store.recordId,
```
RebeccaFrem02, line 18: define   const { recordId } = store in the top of this page and use it as recordId everywhere

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -30,10 +32,70 @@ const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, pr
...
+
+    return response.record
+  }
+
+  const actions = [
+    {
+      key: 'DefaultBilling',
+      condition: true,
+      onClick: setDefaultBilling,
+      disabled: !recordId || store.record.billAddressId == recordId || isDefaultBill
```
RebeccaFrem02, line 88: define all the variables in the store from the begining of the file and use them directly

src/pages/sa-clients/forms/ClientsAddressForm.js
FatimaSalami: i used client id because we already have param named recordId here for the address

src/pages/sa-clients/forms/ClientsAddressForm.js
FatimaSalami: used one time each so no need to defined variable

src/pages/sa-clients/forms/ClientsAddressForm.js
FatimaSalami: usually after posting or closing the record we refetch the form 
so we send new get request and get new status or wip and set it in the form 
so these constants change accordingly 
here i am not sending new get request after setting default buttons.. so i don't have a way to get changed formik values

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -33,7 +37,67 @@ const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, pr
...
```
omarZrayka23, line 90: u can put store.?record.?billAddressId  in state by default

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -33,7 +37,67 @@ const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, pr
...
+    },
+    {
+      key: 'DefaultShipping',
+      condition: true,
+      onClick: setDefaultShipping,
+      disabled: !recordId || store.record.shipAddressId == recordId || isDefaultShip
```
omarZrayka23, line 96: u can put store.record.shipAddressId in state by default

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -4,21 +4,25 @@ import toast from 'react-hot-toast'
...
+  const onSubmit = obj => {
     if (obj) {
       const data = {
         clientId: clientId,
         address: obj,
         addressId: obj.recordId
       }
 
-      await postRequest({
+      postRequest({
```
omarZrayka23, line 25: can u use await instead .then  or put getAddressGridData in .then

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -4,21 +4,25 @@ import toast from 'react-hot-toast'
...
+import { StoreOutlined } from '@mui/icons-material'
 
-const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, props }) => {
+const ClientsAddressForm = ({ getAddressGridData, recordId, window, props, store, setStore }) => {
   const [address, setAddress] = useState()
-  const { postRequest } = useContext(RequestsContext)
+  const { postRequest, getRequest } = useContext(RequestsContext)
   const { platformLabels } = useContext(ControlContext)
+  const [isDefaultShip, setIsDefaultShip] = useState(store?.record?.shipAddressId == recordId)
+  const [isDefaultBill, setIsDefaultBill] = useState(store?.record?.billAddressId == recordId)
```
RebeccaFrem02, line 14: if there two are dynamic why are we still using them as states?

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -4,21 +4,25 @@ import toast from 'react-hot-toast'
 import { ControlContext } from 'src/providers/ControlContext'
 import AddressForm from 'src/components/Shared/AddressForm'
 import { SaleRepository } from 'src/repositories/SaleRepository'
+import { StoreOutlined } from '@mui/icons-material'
 
-const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, props }) => {
+const ClientsAddressForm = ({ getAddressGridData, recordId, window, props, store, setStore }) => {
```
RebeccaFrem02, line 9: there is recordId and store.recordId? 
put them all in the store it is present to reduce the amount of props, and name them ine clientId and one recordId

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -4,21 +4,25 @@ import toast from 'react-hot-toast'
 import { ControlContext } from 'src/providers/ControlContext'
 import AddressForm from 'src/components/Shared/AddressForm'
 import { SaleRepository } from 'src/repositories/SaleRepository'
+import { StoreOutlined } from '@mui/icons-material'
```
RebeccaFrem02, line 7: remove unecessary import

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -33,7 +37,67 @@ const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, pr
...
+    },
+    {
+      key: 'DefaultShipping',
+      condition: true,
+      onClick: setDefaultShipping,
+      disabled: !recordId || isDefaultShip
+    }
+  ]
+
+  return <AddressForm {...{ ...props, address, setAddress, recordId: recordId, onSubmit, actions }} />
```
RebeccaFrem02, line 100: no need for recordId: recordId just pass recordId

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -33,7 +37,67 @@ const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, pr
...
+        ...latestRecord,
+        billAddressId: recordId
+      })
+    })
+
+    setStore(prevStore => ({
+      ...prevStore,
+      record: { ...latestRecord, billAddressId: recordId }
+    }))
+    setIsDefaultBill(true)
```
RebeccaFrem02, line 54: make sure that this is needed

src/pages/sa-clients/forms/ClientsAddressForm.js
```diff
@@ -33,7 +37,67 @@ const ClientsAddressForm = ({ getAddressGridData, clientId, recordId, window, pr
...
+        ...latestRecord,
+        shipAddressId: recordId
+      })
+    })
+
+    setStore(prevStore => ({
+      ...prevStore,
+      record: { ...latestRecord, shipAddressId: recordId }
+    }))
+    setIsDefaultShip(true)
```
RebeccaFrem02, line 72: make sure that this is needed

src/pages/sa-clients/forms/ClientsAddressForm.js
FatimaSalami: i make changes to  make names clearer

omarZrayka23 (approved):

PR #972, opened 2025-02-03
DateSortingFix

Ghinwa-Arab:
AG-Grid (table) wasn't sorting the date correctly. While testing take into consideration that sorting working well for both date and dateTime columns.


imtithalhalalo:
Screenshot 2025-02-03 111003
it is showing 31/01/2025 then 30/01/2025 then 31/01/2025 sort from the middle and skipping the above ones

imtithalhalalo:
Screenshot 2025-02-03 111730
it is not ordering correctly 9 -> 10 -> 9

PR #979, opened 2025-02-04
fixing buttons bug

imtithalhalalo:
fixing buttons bug which shows 
Unhandled Runtime Error
Error: Expected onClick listener to be a function, instead got a value of string type.

### DIFF
--- a/src/components/Shared/FormShell.js
+++ b/src/components/Shared/FormShell.js
@@ -67,319 +67,317 @@ export default function FormShell({
     ? false
     : true
 
-  if (actions) {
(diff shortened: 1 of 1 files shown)


src/components/Shared/FormShell.js
```diff
@@ -67,11 +67,11 @@ export default function FormShell({
     ? false
     : true
 
-  useEffect(() => {
+  if (Array.isArray(actions)) {
```
omarZrayka23, line 70: are u sure need It?

RebeccaFrem02 (approved):

PR #981, opened 2025-02-04
AR-2791

imtithalhalalo:
disable site and remove delete from data grid on post
sales return


omarZrayka23:
Screenshot 2025-02-05 at 11 27 32 AM

i click lock then unlocked => site is disabled

omarZrayka23 (approved):

PR #983, opened 2025-02-05
client relation issuse

omarZrayka23:
client individual list in editMode
Screenshot 2025-02-04 at 1 48 22 PM


FatimaSalami:
some bugs not directly related to your fix but are in the same form
- client ref should be in validation it is required but not giving red color verification on save
- image
why i am getting horizontal scrolling?

FatimaSalami (approved):

PR #1001, opened 2025-02-14
functionId issuse

omarZrayka23:
fixed functionId issuse


FatimaSalami:
in this page api is returning count 3 for number of record list
image

FatimaSalami (approved):

PR #1006, opened 2025-02-18
AR-2865

imtithalhalalo:
adding search for field global popup

### DIFF
--- a/src/components/Shared/FieldGlobalForm.js
+++ b/src/components/Shared/FieldGlobalForm.js
@@ -180,6 +180,16 @@ export default function FieldGlobalForm({ labels, maxAccess, row, invalidate, wi
       )
     : formik.values.gridRows
 
+  function handleRowsChange(newValues) {
+    const updatedRows = formik.values.gridRows.map(row => {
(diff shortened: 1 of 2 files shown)


src/components/Shared/FieldGlobalForm.js
```diff
@@ -164,6 +172,24 @@ export default function FieldGlobalForm({ labels, maxAccess, row, invalidate, wi
     })()
   }, [])
 
+  const filteredData = formik.values.search
```
omarZrayka23, line 175: const { gridRows, search } = formik.values;
const lowerSearch = search.toLowerCase();

const filteredRows = search
  ? gridRows.filter(({ controlId, name }) =>
      controlId?.toString().includes(lowerSearch) || name?.toLowerCase().includes(lowerSearch)
    )
  : gridRows;   u  can simplify ur code

omarZrayka23:
Screenshot 2025-02-18 at 2 04 25 PM

Screenshot 2025-02-18 at 2 04 33 PM

If I write three characters, no result, for example: dtI.

PR #1024, opened 2025-02-20
hide columns fix

FatimaSalami:
the name of maxAccess var was wrong causing the columns not to be hidden from global auth
additional weight column is added in global auth also
phy count item details screen


src/pages/physical-count-itemDetails/index.js
```diff
@@ -49,7 +49,7 @@ const PhysicalCountItemDe = () => {
   }
 
   const { formik } = useForm({
-    maxAccess,
+    access,
```
imtithalhalalo, line 52: here it should be like that maxAccess: access,

PR #1029, opened 2025-02-24
fix live bug readOnly

FatimaSalami:
amount should not be readonly in edit mode


Ghinwa-Arab:
image

if record is processed it should be readOnly

PR #1036, opened 2025-02-28
remove useEffect from every index and put in useForm

omarZrayka23:
In new mode, set the value of dtId that comes from useDocmentType.

pages example: delivery order and purchase invoice


src/hooks/form.js
```diff
@@ -98,5 +99,14 @@ export function useForm({ maxAccess, validate = () => {}, ...formikProps }) {
 
   formik.validationSchema, dynamicValidationSchema(formikProps?.validationSchema)
 
+  const { key, value } = documentType
+
+  useEffect(() => {
+    if (value)
```
imtithalhalalo, line 105: do you need this if (value) if it is being checked in the second one?

PR #1051, opened 2025-03-05
AR-2935 - v5.24

imtithalhalalo:
purity not readOnly and metalId readOnly when there is sku


src/pages/fi-metal-trx/[functionId]/form/MetalTrxFinancialForm.js
```diff
@@ -282,7 +282,10 @@ export default function MetalTrxFinancialForm({ labels, access, recordId, functi
           { from: 'recordId', to: 'metalId' },
           { from: 'purity', to: 'purity' },
           { from: 'purity', to: 'stdPurity' }
-        ]
+        ],
+      },
+      propsReducer({ row, props }) {
+        return { ...props, readOnly: !!row.sku }
```
Comment (NoraBitarSM) [inline src/pages/fi-metal-trx/[functionId]/form/MetalTrxFinancialForm.js:288]: you never check based on the reference of  the field, you always check based on the VALUE
here it's itemId not sku

NoraBitarSM (changes requested):

NoraBitarSM (approved):

PR #1067, opened 2025-03-11
blurIssue-onclear - v5.24 (APPROVED)-merged

Ghinwa-Arab:
In datagrid if a number field column has an onBlur event, attempting to clear the field or click cancel doesn't work, the old value remains stored in formik


omarZrayka23:
Screenshot 2025-03-14 at 9 42 26 AM

the total quantity must be equal 1

imtithalhalalo (approved):

PR #1068, opened 2025-03-11
Negative issue2 - v5.24 (APPROVED)

Ghinwa-Arab:
Test numberfield in a datagrid under the following scenarios:
1. Numbers with decimals (less than 1000)
2. Numbers with thousand separators, including those with decimals
3. Negative numbers, both with and without decimals, and those greater than 1000

Each scenario should be tested in both cases:
-When the system allows viewing leading zeros
-When the system does not allow leading zeros


src/components/Shared/DataGrid/components/numberfield/edit.js
```diff
@@ -49,7 +49,7 @@ export default function NumberfieldEdit({ id, column: { props, field }, value, u
           value: ''
         })
       }}
-      onMouseLeave={() => {
+      onBlur={() => {
         typing.current = false
       }}
       renderMouseLeave={true}
```
FatimaSalami, line 55: renderMouseLeave is the prop changed in CustomNumberField
here you are keeping it and changing onMouseLeave
is it by mistake?

src/components/Shared/DataGrid/index.js
```diff
@@ -678,7 +668,7 @@ export function DataGrid({
 
   const setData = (changes, params) => {
     const id = params.node?.id
-    lastCellStopped.current = ''
+
```
FatimaSalami, line 671: why is this removed? isn't this omar's fix??

FatimaSalami:
Video of 2 bugs sent slack

src/components/Shared/DataGrid/components/numberfield/edit.js
```diff
@@ -49,10 +49,9 @@ export default function NumberfieldEdit({ id, column: { props, field }, value, u
           value: ''
         })
       }}
-      onMouseLeave={() => {
```
FatimaSalami, line 52: add again on mouse leave

RebeccaFrem02 (approved):

FatimaSalami (approved):

PR #1084, opened 2025-03-19
AR-2961

imtithalhalalo:
disabled account reference when no account type is selected 
payment voucher


src/pages/fi-payment-vouchers/forms/FiPaymentVouchersForm.js
```diff
@@ -459,6 +459,7 @@ export default function FiPaymentVouchersForm({ labels, maxAccess: access, recor
                 label={labels.accountReference}
                 valueField='reference'
                 displayField='name'
+                disabled={!formik.values.accountType}
```
NoraBitarSM, line 462: use readOnly instead

NoraBitarSM:
when i delete the account type, account reference should be emptied again

PR #1096, opened 2025-03-24
AR-3005

imtithalhalalo:
new screen verify integrity

### DIFF
--- a/src/components/Shared/SuccessDialog.js
+++ b/src/components/Shared/SuccessDialog.js
@@ -1,0 +1,15 @@
+import ConfirmationDialog from '../ConfirmationDialog'
+
+const SuccessDialog = ({ window, fullScreen, open, message }) => {
+  return (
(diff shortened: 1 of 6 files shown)


src/components/Shared/SuccessDialog.js
```diff
@@ -0,0 +1,15 @@
+import ConfirmationDialog from '../ConfirmationDialog'
+
+const SuccessDialog = ({ window, fullScreen, open, message }) => {
+  return (
+    <ConfirmationDialog
+      open={open[0] ? open[0] : false}
```
Ghinwa-Arab, line 6: Why pass` [true, {}] `to open instead of just open={true}? What's the purpose of the empty object? If there's no reason for it, remove `open[0] ? open[0] : false` and pass open directly

Ghinwa-Arab (approved):

PR #1099, opened 2025-03-25
salesIssue

Ghinwa-Arab:
**The issues were in sales settings:**
-screen doesn't open
-try to clear any combo box and save, value still exists

I have refactored the code and use Contexts for defaults. After saving, please ensure you log out and log back in again. Also, re-test the behavior of check boxes


src/pages/sales-settings/Form/SalesSettingsForm.js
```diff
@@ -36,7 +35,7 @@ export default function SalesSettingsForm({ _labels, access }) {
     'salesTD'
   ]
 
-  const { formik } = useForm({
+  const formik = useFormik({
```
imtithalhalalo, line 38: it can be returned to useForm?

PR #1114, opened 2025-04-10
AR-3025

imtithalhalalo:
header is not showing data for sku and itemName try neg-deploy
site management


FatimaSalami:
i am wondering why we need in initial values to initialize item and sku then in edit mode which is the only mode we set them the same unless we have api response for these fields..

FatimaSalami (approved):

PR #1124, opened 2025-04-11
refetch page after  submit

omarZrayka23:
page name: pm-sketch 
when i save a new sketch, the outer is not refreshed even also if i close the Add/Edit window, i have to click on GO to refresh the outer grid

### DIFF
--- a/src/pages/pm-sketch/index.js
+++ b/src/pages/pm-sketch/index.js
@@ -119,7 +119,8 @@ const Sketch = () => {
     stack({
       Component: SketchForm,
       props: {
-        recordId
+        recordId,
+        invalidate
       },
       width: 700,
       height: 700,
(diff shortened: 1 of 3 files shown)


src/pages/pm-sketch/index.js
```diff
@@ -119,7 +119,8 @@ const Sketch = () => {
     stack({
       Component: SketchForm,
       props: {
-        recordId
+        recordId,
+        invalidate
```
imtithalhalalo, line 123: why you are passing here invalidate? and not using it from the form directly?

omarZrayka23: because used the form of pm-sketch from the second page, and the second page does need to invalidate after submitted.

imtithalhalalo (approved):

PR #1134, opened 2025-04-15
add reourceLookup and fixe in dataGrid onblur

omarZrayka23:
page: pm-model-maker

in this screen, tab Materials, we need to add a new field, SKU, so the grid now will have 4 columns instead of 3, first column SKU, mandatory

also fixed onBlur update  new pieces value


RebeccaFrem02:
image
make the dropdown bigger

PR #1144, opened 2025-04-18
AR-3111

imtithalhalalo:
remove batch number range
system functions


imtithalhalalo (commented):
Screenshot 2025-04-22 110108

PR #1150, opened 2025-04-22
currency  required in generate invoice form

FatimaSalami:
change currency field into required so that they can't generate without having default currency set in pu settings 
to prevent bug
https://softmachine.atlassian.net/browse/AR-2863


imtithalhalalo:
Screenshot 2025-04-23 115031
site is not lightening red and it is required

src/pages/shipments/forms/GenerateInvoiceForm.js
```diff
@@ -52,7 +52,8 @@ export default function GenerateInvoiceForm({ labels, maxAccess: access, recordI
     validationSchema: yup.object({
       plantId: yup.number().required(),
       dtId: yup.number().required(),
-      date: yup.string().required()
+      date: yup.string().required(),
```
imtithalhalalo, line 55: should be date

PR #1169, opened 2025-05-05
AR-3153 worksheetDTD

NoraBitarSM:


src/pages/mf-ws-dtd/Forms/WSDTDForm.js
```diff
@@ -0,0 +1,127 @@
...
+  const { formik } = useForm({
+    initialValues: {
+      recordId,
+      functionId: SystemFunction.Worksheet,
+      dtId: null,
+      workCenterId: null,
+      lineId: null
+    },
+    maxAccess,
+    enableReinitialize: false,
```
imtithalhalalo, line 36: this can be removed because by default it is false

PR #1205, opened 2025-05-20
AR-3236

imtithalhalalo:
adding reference to rowAccessTab users

### DIFF
--- a/src/pages/users/Tabs/RowAccessTab.js
+++ b/src/pages/users/Tabs/RowAccessTab.js
@@ -184,7 +184,11 @@ const RowAccessTab = ({ maxAccess, labels, storeRecordId }) => {
 
   const filteredData = search
     ? {
-        list: data?.list?.filter(item => item?.name && item?.name?.toLowerCase().includes(search.toLowerCase()))
+        list: data?.list?.filter(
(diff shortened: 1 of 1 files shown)


src/pages/users/Tabs/RowAccessTab.js
```diff
@@ -72,10 +77,12 @@ const RowAccessTab = ({ maxAccess, labels, storeRecordId }) => {
     setData({ list: [] })
     classId = classId || ResourceIds.Plants
 
-    const plantRequestPromise = getRequest({
-      extension: SystemRepository.Plant.qry,
-      parameters: '_filter='
-    })
+    const plantRequestPromise =
+      classId == ResourceIds.Plants &&
```
FatimaSalami, line 81: why is this condition added?

FatimaSalami:
search should not search based on reference also now? not only name?

src/pages/users/Tabs/RowAccessTab.js
imtithalhalalo: because it should only be called when ResourceId is Plants it should not be called always

FatimaSalami (approved):

PR #1207, opened 2025-05-20
AR-2970

RebeccaFrem02:
minimize window

### DIFF
--- a/src/components/Shared/Window.js
+++ b/src/components/Shared/Window.js
@@ -2,6 +2,7 @@ import React, { useEffect, useState, useCallback, useMemo, useRef, useContext } 
 import { DialogTitle, DialogContent, Paper, Tabs, Tab, Box, Typography, IconButton } from '@mui/material'
 import ClearIcon from '@mui/icons-material/Clear'
 import OpenInFullIcon from '@mui/icons-material/OpenInFull'
+import MinimizeIcon from '@mui/icons-material/Minimize'
 import Draggable from 'react-draggable'
 import WindowToolbar from './WindowToolbar'
 import { useSettings } from 'src/@core/hooks/useSettings'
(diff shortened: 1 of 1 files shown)


NoraBitarSM:
when i minimize then drag the minimized window, it's like that 
image

NoraBitarSM:
1. when i expand then minimize there is a horizontal scroll that appears 
image


2. the error window minimize's behaviour is different than the other ones 
https://github.com/user-attachments/assets/639b4681-7106-400f-9597-8141dc30fd75

src/pages/bp-master-data/forms/BPMasterDataForm.js
```diff
@@ -56,6 +57,7 @@ export default function BPMasterDataForm({ labels, maxAccess: access, setEditMod
       legalStatusId: null,
       isBlackListed: false
     },
+    enableReinitialize: false,
```
omarZrayka23: clear it , enableReinitialize is false by default

src/components/Shared/Window.js
```diff
@@ -126,20 +152,40 @@ const Window = React.memo(
...
+            >
               <Paper
                 ref={paperRef}
                 tabIndex={-1}
                 sx={{
-                  transition: 'width 0.3s, height 0.3s',
-                  height: controlled ? (expanded ? containerHeight : height) : expanded ? containerHeight : height,
+                  transition:
+                    'max-height 0.35s ease, width 0.35s ease, background-color 0.35s ease, opacity 0.25s ease',
+                  height: !minimized
```
Ghinwa-Arab: This code is very hard to read. can we refactor it to something more readable and cleaner?
U can used this attached refactored code but please check that it won’t cause any issues

```
height: minimized
  ? '40px'
  : expanded
  ? containerHeight
  : height,
```

omarZrayka23:
Screenshot 2025-11-11 at 8 38 01 AM
When maximizing, why window is not take the full tab size?

src/pages/bp-master-data/forms/BPMasterDataForm.js
Ghinwa-Arab: why? we are avoiding adding enableReinitialize whether it is true or false

imtithalhalalo:
image
image
when I have the error there is a scroll to the left but check purchase invoice it has similar window size and it doesn't have this scroll also when I minimize the scroll disappear

PR #1211, opened 2025-05-20
AR-3238

RebeccaFrem02:
plants page:

- add Inactive checkbox in the form
- new filters: Plant Group / Cost Center / Active status (Active/Inactive)
- add new column Plant Group
- add new column Inactive


imtithalhalalo:
Screenshot 2025-05-22 083220
not related but small bug in the screen check the max value

PR #1220, opened 2025-05-23
AR-3254

imtithalhalalo:
modification is receipt voucher

- add Verify button, same as ASP
- change the name of Payment Method combo to Receipt Method.
- add the necessary fields if the Receipt Method is Bank or Transfer or Credit Card (Field Check No enabled and mandatory only if payment method = check), so the details can be registered.
- change the Cash Account field from Lookup to Combo so the user can be able to choose from the drop down list what cash account he wants.
- to show only the Fiat Money currency in Currency field (filter according to currency type)


RebeccaFrem02:
image


put chash account under reciept method and contact under check NO

RebeccaFrem02 (approved):

PR #1223, opened 2025-05-23
AR-3258

imtithalhalalo:
payment voucher outer grid

- add Plant Name column
- add Payment Method column

- when trying to delete a document, we should get the window saying : please enter Delete word to delete.

add filters:

- Payment Method
- Payment Reason (will not be added now)


FatimaSalami:
image
this column is always emtpy

FatimaSalami:
when you press save and form is empty all mandatory fields are red..
when you fill them they are stull red

FatimaSalami (approved):

PR #1227, opened 2025-05-26
AR-3294 - v5.26

imtithalhalalo:
I choose first Yopougon then download it in the second windows, 

second i choose Riviera then I reshuffle, I put Riviera in the first position and Yopougon in the second position.

when I decide to choose another city, it put back as it was initially (it is keeping the initial values Yopougon , Riviera however I sheffled them to Riviera , Yopougon )

Also when I choose a sales zone and drag it to the bottom it is not allowing that

### DIFF
--- a/src/pages/de-gen-trip2/index.js
+++ b/src/pages/de-gen-trip2/index.js
@@ -371,12 +371,18 @@ const GenerateOutboundTransportation2 = () => {
   const handleRowDragEnd = event => {
     if (!event.api) return
 
-    let allNodes = []
-    event.api.forEachNode(node => allNodes.push(node))
+    setTimeout(() => {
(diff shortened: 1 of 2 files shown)


RebeccaFrem02 (approved):

src/pages/de-gen-trip2/index.js
```diff
@@ -371,12 +371,18 @@ const GenerateOutboundTransportation2 = () => {
   const handleRowDragEnd = event => {
     if (!event.api) return
 
-    let allNodes = []
-    event.api.forEachNode(node => allNodes.push(node))
+    setTimeout(() => {
```
omarZrayka23, line 374: put this bhv in table shared component

omarZrayka23 (approved):

PR #1233, opened 2025-05-28
account summary - callAccAPI

Ghinwa-Arab:
account summary


src/components/Shared/AccountSummary.js
```diff
@@ -168,7 +191,6 @@ export default function AccountSummary({ clientInfo, moduleId }) {
                   formik.setFieldValue('agpId', newValue?.recordId || null)
                   if (!newValue?.recordId) setData({ list: [] })
                 }}
-                error={formik?.touched?.agpId && Boolean(formik.errors?.agpId)}
```
imtithalhalalo, line 171: why did you remove the error? is it because no save?

PR #1240, opened 2025-05-29
AR-3309

imtithalhalalo:
users in security groups
user combo should fill from SY.qryUS not 
AU.asmx/qryGUS


omarZrayka23:
Screenshot 2025-05-29 at 12 07 17 PM


add new row  submitted success then close and open window => row added is not founds

PR #1260, opened 2025-06-04
sales inv

Ghinwa-Arab:
getICP should filter according to row's muId


omarZrayka23:
Screenshot 2025-06-05 at 9 13 27 AM

select sky by tab

PR #1261, opened 2025-06-04
AR-3329- v5.27 - approved

imtithalhalalo:
increase width of account lookup dropdown
receipt voucher


Ghinwa-Arab:
image
image

Make sure your changes don’t affect other lookups. On deploy in the sales invoice screen, the items dropdown displays two columns with equal widths but in ur pr not. Could you pls check if we have two columns should they have same width by default?

Ghinwa-Arab (approved):

FatimaSalami (approved):

PR #1274, opened 2025-06-10
AR-3272

imtithalhalalo:
modifications 3d design

- add new field Production Line (this field will be disabled and it will be filled by the document type)
- sketch lookup search result to add 2 columns (Date and Designer)
- sketch lookup should filter by Production Line (to show only the sketches of the production line filled) 
- Designer field should become combo (like sketch screen), and show Type column in result drop down


src/pages/pm-3d-design/forms/ThreeDDesignForm.js
```diff
@@ -264,8 +264,22 @@ export default function ThreeDDesignForm({ labels, access, recordId }) {
...
+                      formik.setFieldValue('productionLineId', null)
+
+                      if (newValue?.recordId) {
+                        const { record } = await getRequest({
+                          extension: ProductModelingRepository.DocumentTypeDefault.get,
+                          parameters: `_dtId=${newValue?.recordId}`
+                        })
+
+                        formik.setFieldValue('productionLineId', record?.productionLineId)
+                      } else {
```
omarZrayka23, line 280: clear else you do not need

omarZrayka23 (approved):

PR #1285, opened 2025-06-12
AR-3272

imtithalhalalo:
productionLine removing required


omarZrayka23:
Screenshot 2025-06-12 at 3 58 33 PM

why when i chenge documenttype with production line => clear value  of  source

omarZrayka23 (approved):

PR #1286, opened 2025-06-12
fixes

omarZrayka23:
add new field Production Line (this field will be disabled and it will be filled by the document type)

Model Maker lookup search result to add 1 column (Date)

Model Maker lookup should filter by Production Line (to show only the Model Maker of the production line filled)

add new field Date

add new field Metal (filled from Model Maker when selected)


imtithalhalalo:
Screenshot 2025-06-13 084916
add date as type

imtithalhalalo (approved):

PR #1289, opened 2025-06-13
AR-3163

imtithalhalalo:
adding date

by default disabled, enable it from global auth
by default show today’s date


NoraBitarSM (changes requested):
delete the date and generaye, you get this error 
image

PR #1317, opened 2025-06-24
AR-3383 AR-3384:  job transfer checkout modifications

FatimaSalami:
add new column Document Type Name
remove wip and itemName columns
qty 2 decimals
image should be set when job order is selected try JO25KN-10148


Ghinwa-Arab:
image
also add decimal scale 2 for qty column

Ghinwa-Arab:
image
it's possible to add max width for the image? because in some cases it's pushing clear button

Ghinwa-Arab (approved):

PR #1325, opened 2025-06-26
fixes-dataGrid-NumberField

omarZrayka23:
https://softmachine.atlassian.net/browse/AR-3415

please check the video in jira
also, Please test the functioning of pages that use the createConditionalSchema method.

### DIFF
--- a/src/components/Inputs/CustomNumberField.js
+++ b/src/components/Inputs/CustomNumberField.js
@@ -97,7 +97,6 @@ const CustomNumberField = ({
       if (inputValue?.length > maxLength) e.target.value = value
     } else {
       if (inputValue?.length > maxLength - (decimalScale || 0)) {
-        console.log(value)
         e.target.value = value
       }
     }


imtithalhalalo:
Screenshot 2025-06-30 084608
when I write value in the field but then focus on it it directly round the number to 100,000,000,000,000.22

imtithalhalalo:
The default value gets cleared when using the cancel button, leaving the cell empty. But when using the keyboard to delete, the default value remains.
Screenshot 2025-06-30 090426

For this point when I clear the value two times it removed the default value is that a correct behaviour? base price for example

Ghinwa-Arab:
image

I entered a large number in the qty field, the volume and weight values in the grid didn’t change, but the totals in the footer show different values

Ghinwa-Arab:
image

on sku selection it updates qty & weight fields by default but in the footer only qty is calculated

imtithalhalalo:
Screenshot 2025-07-01 114242
in the decimal i added .333 but it is adding it as .313 for the field unit price in the data grid

src/components/Inputs/CustomNumberField.js
```diff
@@ -65,6 +65,8 @@ const CustomNumberField = ({
   const handleNumberChangeValue = e => {
     const value = formatNumber(e)
     if (value) e.target.value = value
+
+    console.log(value, parseInputValue(value))
```
imtithalhalalo, line 69: remove log

imtithalhalalo (approved):

PR #1331, opened 2025-06-30
AR-3430

imtithalhalalo:
the reference field is readonly in edit mode in 1 case : if the group is assigned to a number range and it’s not external
otherwise the reference is editable


NoraBitarSM:
i linked the design group ايكلا to a number range 51099114 that is NOT external but the design reference 11NE2867 was still enabled

NoraBitarSM:
edit a record with no group, you will get this error 
image

PR #1335, opened 2025-07-01
Ar 3251

Ghinwa-Arab:
test search behavior well


imtithalhalalo:
Screenshot 2025-07-03 083123
I am searching by qty but it is not showing it

imtithalhalalo (approved):

FatimaSalami:
video about bug deleting from grid when we have value in search is sent

PR #1367, opened 2025-07-10
changes

imtithalhalalo:
when i open the Issue of Materials (add), there are fields that should be filled automatically, its not filled.
check first image in the task https://softmachine.atlassian.net/browse/AR-3381

Expected PCS should be always disabled, like Expected Qty.
the form does not save the data, error on saving.


Ghinwa-Arab:
image
currently, the qty field is accepting only 1 digit and the pcs field is limited to 4 digits. is this on purpose? because in the asp the behavior is different

PR #1373, opened 2025-07-10
AR-3476

imtithalhalalo:
generate purchase transaction


src/pages/pu-trx/[functionId]/index.js
```diff
@@ -24,17 +24,6 @@ const PuTrx = () => {
 
   const { functionId } = Router()
 
-  const getResourceId = functionId => {
```
Comment (NoraBitarSM) [inline src/pages/pu-trx/[functionId]/index.js:27]: aren't you using it in DatasetIdAccess: getResourceId(parseInt(functionId)), ?

PR #1383, opened 2025-07-14
add total qty

omarZrayka23:
in this screen, in tab Materials, we need Total Qty under the grid, column Qty

ex: WS25CN21-12872


imtithalhalalo:
Screenshot 2025-07-15 122341
the qty is in -ve but when you open the form it is in +ve so the total is becoming in -ve

imtithalhalalo (approved):

PR #1388, opened 2025-07-15
fixed value start with .

omarZrayka23:
if i try to write in Qty column decimal number like .5, when i click tab, the number becomes 5 and not 0.5

workSheet example: WS25CN21-13129


imtithalhalalo:
Screenshot 2025-07-15 163225
when I click two times on the (.) it adds zero then the typing will be on the left

imtithalhalalo:
Screenshot 2025-07-17 100308
if I clicked (.) then zero it is removing the (.)
what if I want to enter .01

imtithalhalalo:
when I type 0.0 then delete just the zero it is deleting the (.) I need to re-enter it again

PR #1389, opened 2025-07-16
AR-3380

imtithalhalalo:
issue of materials in worksheet
to have an option to be able to hide Lot column in grid
Total Expected Qty field is showing wrong data.
when entering a new line, all totals under the grid are emptied


Ghinwa-Arab:
image
 i don't know if u have to fix it in this PR or not, but save and clear not working

PR #1413, opened 2025-07-23
AR-3505

imtithalhalalo:
unlock from closing tab of page
now the unlock record works only when the window is closed from X or post or clear.
now we need to have more ways to unlock the record:
when closing all the tab (outer grid)

Fixing Bug in report params 

- end date is not clearing
- it is not taking the time of the laptop
- Please check deploy to reproduce the bug


Ghinwa-Arab:
image
i click clear button, date not clearing

Ghinwa-Arab (approved):

RebeccaFrem02 (approved):

PR #1424, opened 2025-07-28
AR-3376

imtithalhalalo:
Do not allow to save the job order if Expected Pcs is Zero.
after save or before, if i try to change the Routing, it takes the first selection in the lookup.
when you select RT15-16, and click enter, it took the first selection from the combo RT15-10
please try it on deploy to reproduce the bug

On edit mode the title is duplicating when we save


omarZrayka23:
sent video slack

src/components/Inputs/CustomLookup.js
```diff
@@ -118,7 +118,6 @@ const CustomLookup = ({
             setAutoFocus(true)
 
             valueHighlightedOption.current = ''
```
omarZrayka23, line 120: TRY CLEAR IT?

omarZrayka23 (approved):

PR #1429, opened 2025-07-29
add sg combo

FatimaSalami:
add new comboBox sgId

if “is confidential” is checked, then the security group combo is enabled and mandatory
else, it’s readonly, optional and empty


Ghinwa-Arab:
image
image

I added this record, then tried to delete it, but an error occurred. After that, I tried to open the record in edit mode, and the screen showed an unexpected error.

PR #1437, opened 2025-07-31
Vat calculation

NoraBitarSM:


Ghinwa-Arab:
image

item IG-00027 has priceType= weight and taxBase = 2 so vat should be qty* weight which is calculated correctly but on save it's returning this error

FatimaSalami:
image
save button becomes disabled when i add the line

Ghinwa-Arab (approved):

PR #1441, opened 2025-08-01
AR-3587 

FatimaSalami:
when opening the GL popup from the job order screen, the date in the GL should be the “ending date” not the “date”

if endingDate is empty, show an error message and don’t proceed with the popup opening

### DIFF
--- a/src/components/Shared/GeneralLedger.js
+++ b/src/components/Shared/GeneralLedger.js
@@ -142,7 +142,7 @@ const GeneralLedger = ({ functionId, values, valuesPath, datasetId, onReset, win
       if (datasetId === 42604) {
         if (!formValues.endingDT) {
           stackError({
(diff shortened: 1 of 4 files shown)


src/components/Shared/GeneralLedger.js
```diff
@@ -63,7 +65,7 @@ const GeneralLedger = ({ functionId, values, valuesPath, datasetId, onReset, win
     initialValues: {
       recordId: formValues.recordId,
       reference: formValues.reference,
-      date: formValues.date,
+      date: datasetId === 42604 ? formValues.endingDT : formValues.date,
```
omarZrayka23, line 68: why  you are not  from parent componet date: datasetId === 42604 ? formValues.endingDT : formValues.date,

omarZrayka23 (approved):

PR #1447, opened 2025-08-04
AR-3606

omarZrayka23:
- we should submit the form when we click on enter

but make sure the combobox , date  drop down or lookup  doesn’t show an issue when focused on it and we click enter 

- press Ctrl + Save to submit.
- press Escape to close the window.


FatimaSalami:
image
when i get error i can still press enter and save 
 
is this normal behaviour

omarZrayka23:
> image when we press enter ok is selected but then we lose focus in screen so another enter won't save while in asp we still can save after ok is selected

not required

FatimaSalami (approved):

PR #1450, opened 2025-08-05
AR-3601

imtithalhalalo:
reset tv


src/pages/hr-reset-tv/Forms/ResetTVForm.js
```diff
@@ -0,0 +1,144 @@
...
+    initialValues: {
+      startDate: null,
+      endDate: null,
+      employeeId: null,
+      branchId: null
+    },
+    maxAccess,
+    validationSchema: yup.object({
+      startDate: yup.date().required(),
+      endDate: yup.date().required()
```
RebeccaFrem02, line 34: image

i should not be able to select to date that has smaller value than from date

PR #1460, opened 2025-08-06
AR-3614

imtithalhalalo:
fi-open-po


RebeccaFrem02:
image

since there is no add button in this page, is it normal to have clear and save and clear?

PR #1463, opened 2025-08-07
AR-3349

omarZrayka23:
Std Weight column should be read only and it will be filled when the design is selected

Ex Qty = Std weight * Ex PCS and we can make it readonly from global auth

allow saving with no rows

### DIFF
--- a/src/pages/mf-prod-order/Forms/ProductionOrderForm.js
+++ b/src/pages/mf-prod-order/Forms/ProductionOrderForm.js
@@ -27,6 +27,7 @@ import ConfirmationDialog from 'src/components/ConfirmationDialog'
 import { useWindow } from 'src/windows'
 import { SaleRepository } from 'src/repositories/SaleRepository'
 import ImportForm from 'src/components/Shared/ImportForm'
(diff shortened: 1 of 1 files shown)


src/pages/mf-prod-order/Forms/ProductionOrderForm.js
```diff
@@ -118,6 +120,8 @@ export default function ProductionOrderForm({ labels, access, recordId, window }
     }
   })
 
+  console.log(formik.values)
```
imtithalhalalo, line 123: remove log

imtithalhalalo:
Screenshot 2025-08-07 132702
when i save with no rows I cannot add in editMode?

imtithalhalalo:
Screenshot 2025-08-07 135636
Screenshot 2025-08-07 135629
I saved two records after save the grid is empty

imtithalhalalo:
Screenshot 2025-08-07 135636
Screenshot 2025-08-07 135629
when i save two records after save the grid is empty

imtithalhalalo (approved):

PR #1471, opened 2025-08-12
bill of material bug AR-3620

FatimaSalami:
Cost PCT is empty when adding new data in components tab 
adding data should cause recalculation of cost pct of all rows
same for deleting a line

### DIFF
--- a/src/pages/mf-bom/Forms/Component.js
+++ b/src/pages/mf-bom/Forms/Component.js
@@ -82,7 +82,7 @@ const ComponentBOM = ({ store, labels }) => {
     endpointId: ManufacturingRepository.Component.qry
   })
 
-  const totalQty = data?.list.reduce((qtySum, row) => {
+  const totalQty = data?.list?.reduce((qtySum, row) => {
     const qtyValue = parseFloat(row.baseQty) || 0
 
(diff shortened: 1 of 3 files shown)


src/pages/mf-bom/Forms/Component.js
```diff
@@ -108,7 +108,10 @@ const ComponentBOM = ({ store, labels }) => {
         muId: obj?.muId,
         access,
         seqNo: obj?.seqNo ?? maxSeqNo,
-        store
+        store,
+        components: data?.list,
+        invalidate,
```
imtithalhalalo, line 113: why you added invalidate here?

imtithalhalalo (dismissed):

src/pages/mf-bom/Forms/Component.js
```diff
@@ -108,7 +108,9 @@ const ComponentBOM = ({ store, labels }) => {
         muId: obj?.muId,
         access,
         seqNo: obj?.seqNo ?? maxSeqNo,
-        store
+        store,
+        components: data?.list,
```
omarZrayka23, line 112: you do not need pass components and calculateCostPct
you pass function recalculate then Request it after submit

PR #1472, opened 2025-08-12
AR-3600

imtithalhalalo:
i got an error when i tried to change the client name 
from C-0007 (discount zero in client screen) TO  CM-0002 (discount zero in client screen)
neg-deploy
check the image for data in the task https://softmachine.atlassian.net/browse/AR-3600?focusedCommentId=22352

### DIFF
--- a/src/components/Shared/ChangeClient.js
+++ b/src/components/Shared/ChangeClient.js
@@ -59,29 +59,25 @@ export default function ChangeClient({ formValues, onSubmit, window }) {
     }
   })
 
+  function normalize(val) {
+    return val == null ? 0 : val
+  }
+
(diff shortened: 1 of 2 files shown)


src/components/Shared/ChangeClient.js
```diff
@@ -59,6 +59,10 @@ export default function ChangeClient({ formValues, onSubmit, window }) {
     }
   })
 
+  function normalize(val) {
+    return val == null ? 0 : val
```
Ghinwa-Arab, line 63: what if the value is undefined, how this will be handled? isn't better to refactor this condition to ` !val ? 0 : val`

PR #1487, opened 2025-08-19
AR-3612

omarZrayka23:
fixed plantid and siteId

if you notice, Plant and Site are empty, even though i filled them and saved. i closed the window than clicked edit, they disappeared.

i noticed also when you open the add screen, the plant and site are filled, when you select the document type, they are emptied, even though there is no default screen the document type


imtithalhalalo:
const plantId = parseInt(userDefaultsData?.list?.find(obj => obj.key === 'plantId')?.value)
  const siteId = parseInt(userDefaultsData?.list?.find(obj => obj.key === 'siteId')?.value)
  why did you keep these if they are normal fields and will not be filled from defaults or document type?

imtithalhalalo (approved):

PR #1535, opened 2025-09-03
AR-3712

imtithalhalalo:
py settings
new screen


Ghinwa-Arab:
image

they shouldn't accept negative nbrs

src/pages/py-settings/index.js
```diff
@@ -0,0 +1,373 @@
...
+      const data = Object.entries(obj).map(([key, value]) => ({
+        key,
+        value
+      }))
+
+      const res = await postRequest({
+        extension: SystemRepository.Defaults.set,
+        record: JSON.stringify({ sysDefaults: data })
+      })
+      if (res) updateDefaults(data)
```
Ghinwa-Arab, line 103: remove `if (res)`

Ghinwa-Arab:
image
image


You can’t convert the number returned from the api into a positive value on your own. It should be displayed exactly as received. However, since negative numbers aren’t allowed, the field should be highlighted like in ASP.

Ghinwa-Arab (approved):

PR #1556, opened 2025-09-08
AR-3770

omarZrayka23:
migrate page name: sp-categ


src/repositories/RepairRepository.js
```diff
@@ -1,16 +1,24 @@
+import SpCategory from 'src/pages/sp-categ'
+
 const service = 'RS.asmx/'
 
 export const RepairRepository = {
```
imtithalhalalo, line 5: this file has been renamed merge from master and fix the conflict

PR #1565, opened 2025-09-09
Ar 3771

Ghinwa-Arab:
migrate pm template

### DIFF
--- a/src/pages/maint-temp/Forms/TasksForm.js
+++ b/src/pages/maint-temp/Forms/TasksForm.js
@@ -14,11 +14,14 @@ import { RepairAndServiceRepository } from 'src/repositories/RepairAndServiceRep
 import { Grid } from '@mui/material'
 import ResourceComboBox from 'src/components/Shared/ResourceComboBox'
 import { DataSets } from 'src/resources/DataSets'
+import CustomCheckBox from 'src/components/Inputs/CustomCheckBox'
+import CustomNumberField from 'src/components/Inputs/CustomNumberField'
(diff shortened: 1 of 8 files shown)


src/pages/maint-temp/Forms/TasksForm.js
```diff
@@ -0,0 +1,276 @@
...
+          return !tbh || value
+        }),
+      expectedLaborHrs: yup.number().min(1).max(32767).nullable()
+    }),
+    onSubmit: async obj => {
+      await postRequest({
+        extension: RepairAndServiceRepository.MaintenanceTemplateTask.set,
+        record: JSON.stringify(obj)
+      })
+      toast.success(platformLabels.Saved)
```
RebeccaFrem02, line 101: on edit mode it should be updated

Ghinwa-Arab: No, here don’t have a single recordId to determine whether the record is in edit or new mode. Instead, we use two primary keys: taskId (the combo Id) and the main recordId. When saving, the record can either be inserted as new or update an existing one. So, we can’t reliably detect edit mode because I can open a new mode, add a taskId, and it will still be treated as an edit even though it’s actually new.
Note: In ASP, we display the same notification.

RebeccaFrem02: first this is the notification in asp 


image

RebeccaFrem02: second you can use both of the keys to check if saved or updated

RebeccaFrem02 (approved):

PR #1602, opened 2025-09-23
AR-3801

imtithalhalalo:
Add button to Bill of materials so it duplicate the record 
Copy button should be closed and the record to be refreshed to the new one
outer grid should also be refreshed

### DIFF
--- a/src/pages/mf-bom/Forms/CopyForm.js
+++ b/src/pages/mf-bom/Forms/CopyForm.js
@@ -51,7 +51,6 @@ export default function CopyForm({ labels, maxAccess, values, window, setStore, 
           recordId: null
         })
       })
-      toast.success(platformLabels.Added)
       formik.setFieldValue('recordId', response.recordId)
       setStore(prevStore => ({
         ...prevStore,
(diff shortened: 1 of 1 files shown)


src/pages/mf-bom/Forms/BillOfMaterialsForm.js
```diff
@@ -76,21 +79,37 @@ export default function BillOfMaterialsForm({ labels, maxAccess, setStore, store
...
+    })
+  }
+
+  async function onCopy(obj) {
+    stack({
+      Component: CopyForm,
+      props: {
+        labels,
+        maxAccess,
+        recordId: obj.recordId,
```
Ghinwa-Arab, line 102: Remove recordId, since it’s already included in formik.values. Also, remove obj from the onCopy parameters, there’s no need to pass it because formik is a global state

src/pages/mf-bom/Forms/CopyForm.js
```diff
@@ -34,6 +38,11 @@ export default function CopyForm({ labels, maxAccess, values, window, setStore,
       version: yup.string().required()
     }),
     onSubmit: async obj => {
+      const components = await getRequest({
+        extension: ManufacturingRepository.Component.qry,
```
Ghinwa-Arab, line 42: reorder the sequence of actions in this function

Ghinwa-Arab:
image

same issue , i try to add new component for copied record

Ghinwa-Arab (approved):

PR #1603, opened 2025-09-23
fixWidth

Ghinwa-Arab:
When the number of columns is large, they appear compressed. 
Try on Anthony's account, we have a small number of columns, while on the Neg2x account, the number of columns is much larger.

### DIFF
--- a/src/components/Shared/MatrixTable.js
+++ b/src/components/Shared/MatrixTable.js
@@ -70,8 +70,6 @@ const MatrixGrid = ({
                 {`
                   .my-matrix-grid .ag-horizontal-left-spacer,
                   .my-matrix-grid .ag-horizontal-right-spacer {
-                    width: 0px !important;
-                    min-width: 0px !important;
                     overflow-x: hidden !important;
                   }
                 `}


src/components/Shared/MatrixTable.js
```diff
@@ -204,14 +208,15 @@ const MatrixGrid = ({
...
 
   return (
-    <div style={{ height: 450, width: '100%' }} className='ag-theme-alpine'>
+    <div style={{ height: '100%', width: '100%' }} className='ag-theme-alpine'>
       <AgGridReact
         ref={gridRef}
         rowData={initialRowData}
         getRowId={params => params.data.id}
         columnDefs={columnDefs}
         headerHeight={100}
```
RebeccaFrem02, line 217: use headerHeight={'100%'}, it's better

src/components/Shared/MatrixTable.js
```diff
@@ -72,6 +75,7 @@ const MatrixGrid = ({
         headerName: colItem.colLabels,
         sortable: false,
         flex: 1,
+        minWidth: 100,
         headerComponent: props => (
           <div
             style={{
```
RebeccaFrem02, line 81: you can remove transformOrigin: 'center center', if you have allignItems and ustifyContent

PR #1605, opened 2025-09-24
Ar 3831

Ghinwa-Arab:
Cipa-Deploy 
Sales order , Invoice or Return , when we select item for example PATECHBO100 , this item have multiple Measurement with multiple prices , when u select it in sales by default it will give you the default measurement with price  but if you change the mu to another one that have price , the price won’t change

### DIFF
--- a/src/pages/sa-trx/[functionId]/forms/SaleTransactionForm.js
+++ b/src/pages/sa-trx/[functionId]/forms/SaleTransactionForm.js
@@ -623,16 +623,42 @@ export default function SaleTransactionForm({
         ]
       },
       async onChange({ row: { update, newRow } }) {
-        if (newRow) {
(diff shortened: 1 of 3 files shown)


imtithalhalalo:
image
when I clear muRef it shouldn't clear unit price?

imtithalhalalo:
image
when i first open the record CD-002018 and changed the muRef to BOX-24 it changed unit price to 0 but when I reselected BOX-60 now the api is returning 0 
image however it was showing at the beginning 3.00000

imtithalhalalo:
> image when I clear muRef it shouldn't clear unit price?

Ghinwa-Arab:
> image when I clear muRef it shouldn't clear unit price?

no

imtithalhalalo (approved):

PR #1630, opened 2025-10-07
AR-3867

alihajhassn:
in the outer grid, add 2 columns: “is confidential” and “security group”

in the nodes tab, parent combo, show reference and description in the dropdown and on selection

in the ledger tab, rename the columns from seg1 → seg5 instead of seg0 → seg4

in the ledger tab, selected node field, show the reference and description

### DIFF
--- a/src/pages/financial-statements/forms/NodeForm.js
+++ b/src/pages/financial-statements/forms/NodeForm.js
@@ -71,7 +71,6 @@ export default function NodeForm({ labels, maxAccess, mainRecordId, node }) {
           parameters: `_recordId=${node?.current?.nodeId}`
         })
         formik.setValues(res.record)
-        node.current.nodeRef = res?.record?.reference
       }
     })()
   }, [])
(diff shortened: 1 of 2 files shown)


Ghinwa-Arab:
image
image


When I select a row and then add a new node, then switch to the Ledger tab, it only shows the reference of the previously selected row, the description doesn’t appear. The same issue happens when I delete a row (other than the selected one) only the reference is shown.

Ghinwa-Arab:
image
image

after clearing the selected row u should clear 'selected node' field

src/pages/financial-statements/forms/LedgerForm.js
```diff
@@ -18,11 +18,13 @@ import { createConditionalSchema } from 'src/lib/validation'
...
 
 const LedgerForm = ({ node, labels, maxAccess }) => {
-  const { nodeId, nodeRef } = node?.current
+  const { nodeId, nodeRef, nodedesc } = node?.current
+
   const { getRequest, postRequest } = useContext(RequestsContext)
   const { platformLabels } = useContext(ControlContext)
   const { user } = useContext(AuthContext)
   const { stack: stackError } = useError()
+  const data = `${nodeRef || ''}  ${nodedesc || ''}`
```
Ghinwa-Arab, line 27: The data variable is only used for the 'selected node' field, so there’s no need to declare a separate const for it. u can simply pass the concatenated value directly below.

alihajhassn: done

PR #2016, opened 2026-03-12
AR-4429

Ghinwa-Arab:
The language switch is not working. Clicking on french or arabic don’t switch the language it still appears in english


packages/module-auth/src/pages/login/LoginPage.module.css
```diff
@@ -47,6 +47,11 @@
   padding-bottom: 12px;
 }
 
+.language {
```
NoraBitarSM: does this change affect anything ? 
now on deploy the cursor becomes a pointer when hovering on the language

packages/module-auth/src/pages/login/index.js
```diff
@@ -29,11 +31,17 @@ const LinkStyled = styled(Link)(({ theme }) => ({
 const LoginPage = () => {
   const [showPassword, setShowPassword] = useState(false)
   const [errorMessage, setErrorMessage] = useState(null)
+  const { apiPlatformLabels } = useLabelsAccessContext()
```
NoraBitarSM: why did you use this ? 
it's enough to call the api on the click of the language button

PR #2076, opened 2026-04-07
Ar 4508

Ghinwa-Arab:
refactor draft serials transfer


imtithalhalalo:
image
page is not showing data but I've added one record (this is also on deploy) and I tried 0 in search

Ghinwa-Arab:
> image page is not showing data but I've added one record (this is also on deploy) and I tried 0 in search

normal behavior, this is access issue

Ghinwa-Arab:
> image image when i add serial manually it add design ref but when i add it by import button it doesn't add it also when i add it by import and then save the design ref is removed (this is also on deploy)

even on asp like this, i'll not change anything unless it's reported later

imtithalhalalo (approved):

PR #2095, opened 2026-04-13
AR-4522

imtithalhalalo:
mf-scp settings

isInactive is missing in the apis


Ghinwa-Arab:
image

how the name maxLength is greater than reference, pls check if they are correct

imtithalhalalo:
> image
> how the name maxLength is greater than reference, pls check if they are correct

both are 30

Ghinwa-Arab (approved):

PR #2099, opened 2026-04-15
AR-4530

imtithalhalalo:
1- button to invoice inside sales order should be enabled when record is released

2- after clicking to invoice the sales invoice is generated and it should open sales invoice form

3- sales transaction form shared component


NoraBitarSM:
Screenshot 2026-04-15 at 12 27 45 PM

the reference field is editable in new mode but the number range is not external

NoraBitarSM:
Screenshot 2026-04-16 at 12 05 47 PM

s00sr document type has an external number range, the reference was still disabled andoptional, it should have became mandatory and editable

NoraBitarSM (approved):

Ghinwa-Arab:
image

i add new record and click "to invoice", pls ask backend to show clear error so we can know what is missing

Ghinwa-Arab:
image
image

this document type is set as default, but it isn’t active so it shouldn’t be displayed, right? it's currently showing

Ghinwa-Arab (approved):

PR #2108, opened 2026-04-16
AR-4534

Ghinwa-Arab:
- For global auth:
   - to be able hide these fields (Net Serials Weight - Production Class - Standard Class - 3D Design - Rubber)
   - Production Line field - to be able to make it mandatory 
   - to have the chance to make the date field editable
- in SKU lookup: 
   - only Show Finished Goods or Semi Finished Goods 
   - make the lookup dropdown wider
   - name field make it larger
- Samples button:  show 3 decimals for Item Weight, Total Weight & Average fields

### DIFF
--- a/packages/module-manufacturing/src/pages/mf-job-orders/form/JobOrderForm.js
+++ b/packages/module-manufacturing/src/pages/mf-job-orders/form/JobOrderForm.js
@@ -662,10 +662,13 @@ export default function JobOrderForm({
                         name='date'
                         required
                         readOnly
+                        onChange={formik.setFieldValue}
+                        onClear={() => formik.setFieldValue('date', null)}
                         label={labels.date}
                         value={formik?.values?.date}
                         editMode={editMode}
                         maxAccess={maxAccess}
+                        error={formik.touched.date && Boolean(formik.errors.date)}
                       />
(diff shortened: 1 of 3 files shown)


NoraBitarSM:
Screenshot 2026-04-16 at 11 22 23 AM

after hiding 3d design and rubber, the alignment of the fields in the 3rd column is messed up

packages/module-manufacturing/src/pages/mf-job-orders/form/JobOrderForm.js
```diff
@@ -727,6 +730,9 @@ export default function JobOrderForm({
                         endpointId={InventoryRepository.Item.snapshot}
                         name='itemId'
                         readOnly={isCancelled || isPosted}
+                        filter={{
```
NoraBitarSM: use a snapshot for this, never filter a snapshot in the front end, it will show less than the 50 records returned from the backend

ask for a snapshot endpoint that excludes raw materials

NoraBitarSM:
1. make the date field editable, when we want it disabled we can do it from the global auth
2. the alignment is still not fixed when hiding the fields 
image

NoraBitarSM (approved):

PR #2109, opened 2026-04-16
AR-4554

imtithalhalalo:
removing sku from design


NoraBitarSM:
the fields should be removed from the backend also
itemId
sku
itemName

PR #2115, opened 2026-04-17
AR-4535

imtithalhalalo:
- in Semi Finished Item field in header, after save, we cannot edit the field, its disabled (allow edit)
- add Production Line field in header, will be filled automatically when selecting the Job Order
- Average Weight field make it 3 decimals
- in grid:
           - add new column To Site, at the end of the grid, mandatory
- to add 2 new fields (From Semi Finished Site and To Finished Goods Site)

### DIFF
--- a/packages/module-manufacturing/src/pages/mf-jo-wizard/Forms/JobOrderWizardForm.js
+++ b/packages/module-manufacturing/src/pages/mf-jo-wizard/Forms/JobOrderWizardForm.js
@@ -103,7 +103,8 @@ export default function JobOrderWizardForm({ labels, access, recordId }) {
             sku: yup.string().required(),
             issued: yup.number().required(),
             returned: yup.number().required(),
-            consumed: yup.number().required()
+            consumed: yup.number().required(),
+            toSiteId: yup.number().required(),
           })
         )
         .required()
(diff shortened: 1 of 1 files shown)


Ghinwa-Arab:
image

production line fills from the job order, should it be readOnly or editable like it's so the user can change the value?

packages/module-manufacturing/src/pages/mf-jo-wizard/Forms/JobOrderWizardForm.js
```diff
@@ -82,7 +82,9 @@ export default function JobOrderWizardForm({ labels, access, recordId }) {
           issued: 0,
           returned: 0,
           consumed: 0,
-          seqNo: 1
+          seqNo: 1,
+          siteId: null,
+          siteName: ''
```
Ghinwa-Arab: siteName not necessary to be added in initial values

packages/module-manufacturing/src/pages/mf-jo-wizard/Forms/JobOrderWizardForm.js
```diff
@@ -582,6 +591,84 @@ export default function JobOrderWizardForm({ labels, access, recordId }) {
...
+                }}
+                error={formik.touched.header?.lineId && Boolean(formik.errors.header?.lineId)}
+              />
+            </Grid>
+            <Grid item xs={4}>
+              <ResourceComboBox
+                endpointId={ManufacturingRepository.ProductionShifts.qry}
+                name='header.shiftId'
+                label={labels.shift}
+                maxAccess={access}
```
Ghinwa-Arab: should be maxAccess not access

imtithalhalalo:
> image
> production line fills from the job order, should it be readOnly or editable like it's so the user can change the value?

yes editable

Ghinwa-Arab (approved):

PR #2117, opened 2026-04-17
AR-4505

imtithalhalalo:
on close, we will check for the mismatch of the metal grid.total weight and the header.total weight

we will show a message to “recalculate”

we will add a button “recalculate” it can be used anytime if the record is editable


NoraBitarSM:
Screenshot 2026-04-20 at 10 57 14 AM

click edit to open any record, you get this error

PR #2127, opened 2026-04-21
AR-4549

Ghinwa-Arab:
employee list screen -> Migrate employee termination popup


imtithalhalalo:
on asp after termination it closes the employee and the form not sure if you need to do the same
(but i think to keep employee form opened is better)

imtithalhalalo (approved):

PR #2132, opened 2026-04-22
AR-4555

imtithalhalalo:
job order on item selection
On Item selection
from SKU:

- Production Line
- Routing
- design
- Item Category

when filling design from this Design

- Production Class
- Production Standard
- Picture according to settings from item or from design (design was missing in the fill item function)
- STD Item Weight

and the Work Center should be filled from Routing(first work center in routing)


Ghinwa-Arab:
image

when we clear the item it's clearing the design too, all fields that fills from design are cleared except rubber and 3d design

Ghinwa-Arab (approved):

PR #2136, opened 2026-04-23
AR-4553

Ghinwa-Arab:
- Migrate Employee Missing Details
- all 'employee list' screen tabs now are shared forms under EmployeePages folder


packages/module-hr/src/pages/employee-missing-details/Form/EmployeeMissingList.js
```diff
@@ -0,0 +1,128 @@
...
+  const columns = [
+    {
+      field: 'pictureUrl',
+      headerName: '',
+      flex: 1,
+      type: 'image'
+    },
+    {
+      field: 'parent.reference',
+      headerName: labels.ref,
```
NoraBitarSM, line 49: label value should be reference instead of parent reference

packages/module-hr/src/pages/employee-missing-details/Form/EmployeeMissingList.js
```diff
@@ -0,0 +1,128 @@
...
+      flex: 1
+    },
+    {
+      field: 'parent.fullName',
+      headerName: labels.name,
```
NoraBitarSM, line 54: label value should be full name instead of parent full name

packages/module-hr/src/pages/employee-missing-details/Form/EmployeeMissingList.js
```diff
@@ -0,0 +1,128 @@
...
+      flex: 2,
+      wrapText: true,
+      autoHeight: true
+    },
+    {
+      field: 'department.name',
+      headerName: labels.department,
```
NoraBitarSM, line 61: department not department name
same for the rest of the columns

NoraBitarSM:
Screenshot 2026-04-24 at 12 07 40 PM

employee window isn't showing a title when opened from employee missing details

NoraBitarSM:
terminate an employee from the employee missing details popup, the window remains open and the fields and save remain editable

packages/module-hr/src/pages/employee-missing-details/Form/EmployeeMissingList.js
```diff
@@ -0,0 +1,128 @@
...
+    queryFn: fetchGridData,
+    endpointId: ReportRepository.EmployeeMissingDetails.RT107b,
+    datasetId: ResourceIds.EmployeeMissingDetails,
+  })
+
+  const {
+    labels: employeeLabels
+  } = useResourceQuery({
+    queryFn: fetchGridData,
+    datasetId: ResourceIds.EmployeeFilter
```
NoraBitarSM: this shouldn't be here since the employee tabs are shared now
the maxAccess and qryLBL are requested each time we open the employee missing details window even when we do not open the employee details

packages/module-hr/src/pages/employee-missing-details/Form/EmployeeMissingList.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  return (
+    <VertLayout>
+      <Grow>
+        <Table
+          name='Employeetable'
+          columns={columns}
+          gridData={data}
+          rowId={['recordId']}
```
NoraBitarSM, line 116: /RT.asmx/RT107b?_fieldId=1&_activeStatus=0 doesn't return a recordId
you should use the value from hireInfo.employeeId or parent.recordId

packages/shared-ui/src/components/Shared/EmployeeListWindow.js
NoraBitarSM, line 1: make it a shared component and use if from both screens employee missing details and employee list

packages/module-hr/src/pages/employee-missing-details/index.js
```diff
@@ -0,0 +1,140 @@
...
+            </Grid>
+          }
+        />
+      </Fixed>
+      <Grow>
+        <Table
+          name='table'
+          columns={columns}
+          gridData={ {list: data }}
+          rowId={['recordId']}
```
NoraBitarSM: RT.asmx/RT107?_activeStatus=0 doesn't return a recordId, 
it returns a fieldId

NoraBitarSM:
Screenshot 2026-04-24 at 5 09 57 PM

still no title

packages/shared-ui/src/components/Shared/EmployeePages/ProfileForm.js
```diff
@@ -58,7 +58,7 @@ const ProfileForm = ({ labels, maxAccess, setStore, store, imageUploadRef, getDa
       placeOfBirth: '',
       hireDate: null,
       gender: 1,
-      activeStatus: 1,
+      activeStatus: activeStatus || true,
```
NoraBitarSM: if you're working with a boolean activeStatus it means that it has only true or false, i prefer if you name it something else related to a boolean value or if you work on enum values like any status field in Argus

NoraBitarSM:
Screenshot 2026-04-27 at 4 10 41 PM

employee terminated, the + button should be hidden or disabled in this case
if the edit button is hidden then how can we see the details of the form ? 

same in the remaining tabs

packages/module-hr/src/pages/employee-missing-details/index.js
```diff
@@ -0,0 +1,140 @@
...
+                />
+            </Grid>
+          }
+        />
+      </Fixed>
+      <Grow>
+        <Table
+          name='table'
+          columns={columns}
+          gridData={ {list: data }}
```
imtithalhalalo: the formatting should be {{ list: data }}

packages/shared-ui/src/components/Shared/EmployeeListWindow.js
```diff
@@ -2,25 +2,25 @@ import { Grid, Box, Typography } from '@mui/material'
...
+import SkillsTab from '@argus/shared-ui/src/components/Shared/EmployeePages/SkillsTab'
+import UserDefinedTab from '@argus/shared-ui/src/components/Shared/EmployeePages/UserDefinedTab'
+
+export default function EmployeeListWindow ({ recordId, labels, maxAccess, employeeStatus, window}) {
   const [activeTab, setActiveTab] = useState(0)
   const [store, setStore] = useState({ recordId: recordId || null, hireDate: null })
   const { getRequest } = useContext(RequestsContext)
   const [quickView, setQuickView] = useState(null)
-
+  const isActive = employeeStatus ? employeeStatus == 1 : true
```
imtithalhalalo: here you want only active is status is 1? if yes it can be just like that 
const isActive = employeeStatus == 1

imtithalhalalo:
image
on asp when i clear status i can clear it and it shows all of them zero
but react it keeps it as All selection

imtithalhalalo:
image
in asp it shows 56 record for employment status but in react when i open it is shows 0 records although it has count 56

imtithalhalalo:
image
image

i removed termination from this record all the other tabs became enabled however react it stays disabled for the other tabs only first tab became enabled

packages/shared-ui/src/components/Shared/EmployeeListWindow.js
Ghinwa-Arab: In new mode, employeeStatus is undefined, so comparing it to 1 returns false, which makes isActive false. As a result, all fields become readOnly. To avoid this, I first check whether employeeStatus is defined and return true by default when it isn’t.

imtithalhalalo (approved):

NoraBitarSM:
there is a glitch that happens in the tables when you open popups in employee missing details
this glitch doesn't happen in job order for example even if i open multiple popups on top of each other (job order, worksheet, issue of materials)
it doesn't happen also in the employee list form so it's not related to the employee window

imtithalhalalo (approved):

PR #2139, opened 2026-04-24
AR-4526

Ghinwa-Arab:
remove Quick Search + change ReportParam Name to SADTD_2 for sales return doc type default and SADTD_1 for invoice doc type default


packages/module-sales/src/pages/sa-trx-dtd/[functionId]/index.js
Comment (NoraBitarSM) [inline packages/module-sales/src/pages/sa-trx-dtd/[functionId]/index.js]: you don;t need this anymore

PR #2147, opened 2026-04-28
AR-4571

imtithalhalalo:
users ai tab
when saving the keys 
the stored values should be the values  of combos not the  keys  :
example
AI_provider_Id → openAI
AI_Model_Version → gpt-4o (not  1)

### DIFF
--- a/packages/module-auth/src/pages/users/Tabs/AIForm.js
+++ b/packages/module-auth/src/pages/users/Tabs/AIForm.js
@@ -89,9 +89,11 @@ const AIForm = ({ labels, maxAccess, storeRecordId }) => {
                 required
                 maxAccess={maxAccess}
                 onChange={(_, newValue) => {
-                  formik.setFieldValue('AI_Model_Version', null)
-                  
-                  formik.setFieldValue('AI_provider_Id', newValue?.value || '')
+                  formik.setValues({
+                    ...formik.values,
+                    AI_provider_Id: newValue?.value || '',
+                    AI_Model_Version: ''
+                  })
                 }}
                 error={formik.touched.AI_provider_Id && Boolean(formik.errors.AI_provider_Id)}
               />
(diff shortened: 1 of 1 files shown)


packages/module-auth/src/pages/users/Tabs/AIForm.js
```diff
@@ -13,21 +13,36 @@ import Form from '@argus/shared-ui/src/components/Shared/Form'
...
+          else reject()
+        }
+      })
+    })
+  }
 
   const { formik } = useForm({
     initialValues: {
       AI_provider_Id: null,
+      AI_provider_Id_key: null,
```
Ghinwa-Arab: i see it's not necessary to have this field in the formik, usually formik have only the fields need to be sent in the payload

packages/module-auth/src/pages/users/Tabs/AIForm.js
```diff
@@ -86,13 +84,14 @@ const AIForm = ({ labels, maxAccess, storeRecordId }) => {
...
-                valueField='key'
+                valueField='value'
                 displayField='value'
                 required
                 maxAccess={maxAccess}
                 onChange={(_, newValue) => {
-                  formik.setFieldValue('AI_provider_Id', newValue?.key || null)
                   formik.setFieldValue('AI_Model_Version', null)
+                  
+                  formik.setFieldValue('AI_provider_Id', newValue?.value || '')
```
Ghinwa-Arab: Image

if I press enter to trigger validation then fill out the form & clear only the first combo, only the first combo is marked as required (red) ->this issue can be fixed using setValues instead setFieldValue separately

Ghinwa-Arab:
image

account: anthonys
when I open this user (Argus Automated BackendAABES Services) and immediately save the existing data, it saves successfully even though the first required field is empty

Ghinwa-Arab (approved):

PR #2159, opened 2026-05-05
AR-4610

Ghinwa-Arab:
MF/designs

the reference field should be editable and mandatory only if
- there is no design group
- the design group is not linked to a number range
- the number range is external


NoraBitarSM:
the field can be editable only in the case of an external number range or if there is no number range

imtithalhalalo (approved):

PR #2167, opened 2026-05-07
AR-4621 - merged

Ghinwa-Arab:
### **---> Task AR-4621:**  
Change useRefBehavior to useFieldBehavior for screens that don’t use document type but still require field behavior based on nraId. Then refactor useFieldBehavior structure to match the useRefBehavior structure when passed to useForm which creates a unified prop pattern for both behaviors & inside useForm we handle this prop properly for both cases.

1. useFieldBehavior exists in the following screens:
-master data
-items
-designs
-vendors
-clients

2. new useRefBehavior passing structure is applied
ex: instead of passing this structure to the formik
` documentType: { key: 'dtId', value: documentType?.dtId },`
now we are passing
`  behavior: { key: 'dtId', value: documentType?.dtId, fieldBehavior: documentType?.reference },`

### **---> Task AR-4624:** 
Do not use `changeDT` from `useDocBehavior` together with any other function inside the `onChange` event of the `dtId` combo.

2. Some screens currently depend on `documentType.dtId` inside the `useEffect` dependency array to load default fields, which is incorrect.

3. Pass over all document type combos to ensure only active document types are shown

**Caught scenarios:**
1. When the document type is set by default, the related fields that should load based on that document type are not being populated in the following screens:
* PM Casting
* PM Model Maker
* Rubber
* 3D Drawing

2. The default document type is not displayed at all in:
* PU Draft Serials Return


packages/module-sales/src/pages/sa-clients/forms/ClientsForm.js
NoraBitarSM, line 1: Image

when selecting again a group with non external NRA, empty the reference field

check the rest of the pages if we have the same issue please

packages/module-delivery/src/pages/outbound-transportation/forms/OutboundTranspForm.js
NoraBitarSM, line 1: Image

Anthonys
select a document type with no number range, the reference was still readonly and optional

Image

packages/module-purchase/src/pages/pu-vendors/form/VendorsForm.js
```diff
@@ -139,9 +137,9 @@ export default function VendorsForm({ labels, maxAccess: access, recordId, setSt
                   { key: 'name', value: 'Name' }
                 ]}
                 values={formik.values}
-                onChange={(event, newValue) => {
+                onChange={(_, newValue) => {
                   formik.setFieldValue('groupId', newValue ? newValue.recordId : '')
```
imtithalhalalo: || null

packages/module-sales/src/pages/sa-clients/forms/ClientsForm.js
```diff
@@ -149,17 +165,16 @@ const ClientsForms = ({ labels, maxAccess: access, setStore, store }) => {
                 valueField='recordId'
                 displayField='name'
                 values={formik.values}
-                onChange={(event, newValue) => {
+                onChange={(_, newValue) => {
                   formik.setFieldValue('cgId', newValue?.recordId || '')
```
imtithalhalalo: || null

NoraBitarSM:
please merge from master and do the changes in the new fixing activity

NoraBitarSM:
Screenshot 2026-05-26 at 9 53 54 AM

purchase invoice in edit mode when there is no document type
PU.asmx/getDTD?_dtId=null

i didn't check if this bug is present in other screens

NoraBitarSM:
Anthonys account
ex: sales return:
the system function is linked to an external NRA, so if no doc type is selected, the reference is editable and mandatory which is correct

bug scenario : 
set a user default document type that has a number range that is not external (ex ROI Return on Invoice)
in the activity click add, the doc type is set by default and the reference is rradonly and optional, which is also correct
delete the dId selection
the reference field is still disabled

Ghinwa-Arab:
> Screenshot 2026-05-26 at 9 53 54 AM
> purchase invoice in edit mode when there is no document type PU.asmx/getDTD?_dtId=null
> 
> i didn't check if this bug is present in other screens

this issue only present in this screen

imtithalhalalo:
image
reference has value but still red
same issue in purchase invoice

imtithalhalalo:
image
document type has number range not external and it is editable and optional

imtithalhalalo:
image
in editMode document type enabled

packages/shared-ui/src/components/Shared/Forms/MaterialRequestForm.js
NoraBitarSM, line 1: Image
dtId enabled in edit mode

NoraBitarSM:
Screenshot 2026-06-09 at 8 20 29 AM

dtId enabled in edit mode

packages/shared-ui/src/components/Shared/Forms/SalesOrderForm.js
NoraBitarSM, line 1: Image
dtId editable in edit mode

Ghinwa-Arab: it has 'force enabled' access

PR #2171, opened 2026-05-08
AR-4636

imtithalhalalo:
checkbox


Ghinwa-Arab:
image
image

this issue is present on deploy, but I’m not sure if it needs to be fixed. When I manually check all rows, the 'Check All' option remains unchecked. Also, when I use 'Check All' and then uncheck one record, it still shows as if all records are selected

Ghinwa-Arab:
image

using search the checkbox needs double click to be unchecked (search for checked row then try to uncheck it)

Ghinwa-Arab (approved):

PR #2175, opened 2026-05-12
item category change

imtithalhalalo:
on item category change
- no tables should be emptied and the categ vol column should be updated, total category volume field and total cat vol pct should be updated with preserving the checkboxes in the first table (sales zones) 
sales zone abidjan try changing between PH6 and PH6" for item category try cil3


Ghinwa-Arab (approved):

NoraBitarSM:
the sales zone table scrolls again to the top when i do any change to the category combo, can you prevent this like the table in any outer grid ? when we refresh the data it doesn't scroll again to the top

PR #2178, opened 2026-05-13
AR-4650

imtithalhalalo:
currencies
adding base currency combo
when currency type is metal then the base currency is enabled and optional otherwise readOnly
the combo shows the currencies that are metal only


NoraBitarSM:
Screenshot 2026-05-13 at 11 58 25 AM

1. base currency should show only metal currencies
"the combo shows the currencies that are metal only" this was mentioned in the task description

2. when you select the type "fiat money" the base currency selection is still filled in the combo

PR #2180, opened 2026-05-13
AR-4651

imtithalhalalo:
new tab currencies

this tab shows a table with all the currencies with a checkbox as a first column
qryACU for the checked currencies


packages/module-financials/src/pages/fi-accounts/forms/CurrenciesTable.js
```diff
@@ -0,0 +1,96 @@
...
+  ]
+
+  return (
+    <Form onSave={handleSubmit} maxAccess={maxAccess} fullSize>
+      <VertLayout>
+        <Grow>
+          <Table
+            columns={columns}
+            gridData={{ list: data }}
+            rowId={['recordId']}
```
NoraBitarSM: the mergedData contains a currencyId not a recordId

PR #2181, opened 2026-05-14
Performance

NoraBitarSM:
changes to shared components to reduce errors that are showing in the console


NoraBitarSM:
change in most files 


MUI says:

If you use xs, sm, md, etc., the Grid must also be an item.

So change them to:

<Grid container item spacing={2} xs={8}>

and

<Grid container item xs={4} sx={{ pl: 2 }}>
image

PR #2183, opened 2026-05-14
AR-4653

imtithalhalalo:
wrapping text issue
you can try text wrap in
employee missing details (image + text wrap height)
sms request logs no image + text wrap height (cil3) Delivery as module and trip as resourceId
employee list also image and wrapping text


NoraBitarSM:
Screenshot 2026-05-14 at 10 39 28 AM

when i decrease the width of the wrap text column, it should increase the height of the row to wrap text

PR #2184, opened 2026-05-14
AR-4673 LIVE - merged

Ghinwa-Arab:
- In IV Settings: add new combo 'dimension group'
- In Items:  show this default if there's no item group selected or if the selected item group has no dimension grp

### DIFF
--- a/packages/module-inventory/src/pages/iv-items/forms/ItemsForm.js
+++ b/packages/module-inventory/src/pages/iv-items/forms/ItemsForm.js
@@ -40,6 +40,7 @@ export default function ItemsForm({ labels, maxAccess: access, setStore, store, 
 
   const currencyId = systemDefaults?.list?.find(({ key }) => key === 'currencyId')?.value
   const plId = systemDefaults?.list?.find(({ key }) => key === 'plId')?.value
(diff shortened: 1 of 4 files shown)


packages/module-inventory/src/pages/iv-items/forms/ItemsForm.js
```diff
@@ -200,6 +185,9 @@ export default function ItemsForm({ labels, maxAccess: access, setStore, store,
     formik.setValues({ ...res.record, kitItem: !!res.record.kitItem, isExternal })
     setShowLotCategories(res.record.trackBy === 2)
     setShowSerialProfiles(res.record.trackBy === 1)
+
+    let dimensionGrpName = ''
+    if(!res.record?.dmgId && dgId) dimensionGrpName = await getDimensionGrpName()
```
NoraBitarSM: do you need this ? 
getIT returns always dmgName and getIT is always requested after setIT

Ghinwa-Arab: I only call getDimensionName when getIT does not return  dmgId. In that case, I set the default value and then call getDIG to retrieve the corresponding name

PR #2192, opened 2026-05-15
AR-4681

NoraBitarSM:
added module_bt new module
bt-defaults new page with measurement schedule optional combo

### DIFF
--- a/packages/module-bt/src/pages/bt-defaults/form/BtDefaults.js
+++ b/packages/module-bt/src/pages/bt-defaults/form/BtDefaults.js
@@ -1,0 +1,85 @@
+import { useEffect, useContext } from 'react'
+import { Grid } from '@mui/material'
+import toast from 'react-hot-toast'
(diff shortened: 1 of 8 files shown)


packages/module-bt/src/pages/bt-defaults/form/BtDefaults.js
```diff
@@ -0,0 +1,85 @@
...
+                label={_labels.measurementSchedule}
+                valueField='recordId'
+                displayField={['reference', 'name']}
+                columnsInDropDown={[
+                  { key: 'reference', value: 'Reference' },
+                  { key: 'name', value: 'Name' }
+                ]}
+                values={formik.values}
+                onChange={(_, newValue) => {
+                  formik.setFieldValue('fixing_msId', newValue?.recordId || '')
```
Ghinwa-Arab, line 73: since it’s id, it’s better to reset it to null instead of an empty string

Ghinwa-Arab (approved):

PR #2195, opened 2026-05-15
AR-4606

NoraBitarSM:
FS account details popup
we added a button for the rows that are children and not the subtotal row
the button will open a table showing fixed columns and dynamic columns with total fields

fixed also a bug in numberfield helper that was showing the decimal point without numbers when sending decimal = 0 for a number column
ex: it was showing 23. instead of 23
please test also with the show leading zeros system check


Ghinwa-Arab:
image
image


When I manually sum the total for EGP, Fiat & G21 , the result is +1 compared to the displayed value

Ghinwa-Arab:
image
image

Issues present on deploy also:
- why for 81,905,246- the negative is displayed on the right of the nb?
- if u scroll down, u will see negative numbers are displayed with brackets instead of -ve sign, this was done on purpose?
- on print there's no scroll to view the rest of the report

Ghinwa-Arab (approved):

imtithalhalalo (approved):

PR #2206, opened 2026-05-19
AR-4675 Live merged

Ghinwa-Arab:
3D Print screen:
1. make the production line mandatory
2. make the 3D drawing readonly until we have a production line


NoraBitarSM:
when i change the doc type, production line changes also and 3d drawing is keeping the selection, 
so the current selection does not belong to the new production line

PR #2216, opened 2026-05-22
AR-4705

imtithalhalalo:
hr-leave-types


Ghinwa-Arab:
image

when isPaid is check don't show deduction as required

PR #2217, opened 2026-05-22
AR-4700

Ghinwa-Arab:
migrate change vendor btn to purchase inv


imtithalhalalo:
image
image
i chose PV-00146 as change to vendor react is showing different error than asp is that on purpose?

imtithalhalalo:
image
payment method is empty in the form

imtithalhalalo:
when error Tax Mismatch, Vat mismatch shows the change to is not being cleared however first time it is clearing it


image

imtithalhalalo (approved):

PR #2219, opened 2026-05-22
AR-4684

imtithalhalalo:
- add Attachment button like ASP
- add option to be able to hide column Source Reference
- Cost Center Ref and Cost Center name should be combined into 1 column, and the column should be a Combo (Ref + Name) with search, and when selected show Ref + Name
- Rename Third Party Ref to TPRef


Ghinwa-Arab:
image

label it cost center only, since it shows both ref and name not only ref

Ghinwa-Arab:
image

maximize the width of drop down

Ghinwa-Arab (approved):

PR #2221, opened 2026-05-25
AR-4715 Live

imtithalhalalo:
implementing the “disable sku lookup” feature in materials transfer
based on the ‘disable sku lookup’ value from the document type selection
Adding jump to next line
validation header before filling datagrid


NoraBitarSM:
since the grid sku field is dependent on the document type, you need to disable dtId when there are items in the list

NoraBitarSM (approved):

PR #2230, opened 2026-05-26
AR-4709

imtithalhalalo:
receipt voucher button in sales invoice should open the form in receipt vouchers screen 
ClientId should get the account from getCL api and fill account ref and account name in receipt voucher form
Currency
Amount
Description field will be filled with the Invoice Number

these fields should be filled automatically from the sales invoice


NoraBitarSM:
add the resourceId to the commonResourceIds to prevent requesting maxAccess each time since it's now a shared form

NoraBitarSM:
Screenshot 2026-05-26 at 2 54 01 PM

receipt method is a mandatory field, it didn't lighten red when saving with an empty value

NoraBitarSM:
Screenshot 2026-05-26 at 2 55 17 PM

error when saving
MC.asmx/getDRI was not requested when opening the RV form

packages/shared-ui/src/components/Shared/Forms/SaleTransactionForm.js
```diff
@@ -1240,6 +1250,12 @@ export default function SaleTransactionForm({
       condition: true,
       onClick: onCommissionClick,
       disabled: !isPosted
+    },
+    {
+      key: 'RV',
+      condition: true,
+      onClick: onReceiptVoucher,
+      disabled: !editMode
```
NoraBitarSM: please make the button enabled only when the sales invoice is posted, i just checked with george

PR #2237, opened 2026-06-01
AR-4721

Ghinwa-Arab:
On item selection isVatable per item is set as true if it has a tax schedule
then if IsVatable the Tax combo will be enabled always and i can select or empty the selection & if not IsVatable the Tax combo will always be disabled


NoraBitarSM:
in both new and edit mode, enable the tax schedule combo only if the invoice is vatable and the item is vatable

imtithalhalalo (dismissed):

imtithalhalalo (approved):

imtithalhalalo:
image
when changing in the tax details combo and you have tax in the header it shows this error

imtithalhalalo (approved):

PR #2243, opened 2026-06-02
AR-4660 & AR-4661

Ghinwa-Arab:
In financials receiptVoucher & paymentVoucher:
add new field called balance that shows cash account balance where this field is readOnly that loads from getACB & in edit mode it's returned in the get api

### DIFF
--- a/packages/shared-ui/src/components/Shared/Forms/FIReceiptVoucherForm.js
+++ b/packages/shared-ui/src/components/Shared/Forms/FIReceiptVoucherForm.js
@@ -206,6 +206,17 @@ export default function FIReceiptVoucherForm({ header, recordId, window }) {
     formik.setFieldValue('cashAccountName', cashAccountResult?.name || '')
   }
 
+  async function getBalance (accId, currencyId) {
+    if (!accId || !currencyId) return 
+
+    const res = await getRequest({
+      extension: FinancialRepository.AccountCreditBalance.get,
+      parameters: `_accountId=${accId}&_currencyId=${currencyId}`
+    })
+
+    return res?.record?.balance
+  }
+
   useEffect(() => {
     ;(async function () {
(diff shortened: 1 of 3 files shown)


packages/shared-ui/src/components/Shared/Forms/FIReceiptVoucherForm.js
```diff
@@ -206,6 +206,21 @@ export default function FIReceiptVoucherForm({ header, recordId, window }) {
...
+      formik.setFieldValue('accountBalance', 0)
+      
+      return
+    }
+
+    const res = await getRequest({
+      extension: FinancialRepository.AccountCreditBalance.get,
+      parameters: `_accountId=${accId}&_currencyId=${currencyId}`
+    })
+
```
NoraBitarSM, line 216: return res?.record?.balance

packages/shared-ui/src/components/Shared/Forms/FIReceiptVoucherForm.js
```diff
@@ -625,11 +644,21 @@ export default function FIReceiptVoucherForm({ header, recordId, window }) {
                       await getMultiCurrencyFormData(newValue?.recordId, formik.values.date)
                       formik.setFieldValue('currencyId', newValue?.recordId)
                       formik.setFieldValue('currencyName', newValue?.name)
+                      getBalance(formik.values.cashAccountId, newValue?.recordId)
```
NoraBitarSM: naming the function 'get balance' means that it will only get the balance, and you use the returned result  the way you want 
in this case if you want to keep the naming, you set it here

const balance = await getBalance(formik.values.cashAccountId, newValue?.recordId)
formik.setFieldValue('balance', balance || 0)

NoraBitarSM (approved):

PR #2244, opened 2026-06-02
AR-4756

Ghinwa-Arab:
Bank account screen:
AccountNo should accept characters also


imtithalhalalo:
image
not related but when i write name for account it is directly filling the name of account reference (on deploy)

imtithalhalalo (approved):

PR #2247, opened 2026-06-03
AR-4733

imtithalhalalo:
exchange rate grid in exchange table screen
in exchange table screen , add a new tab for exchange rates,
the tab is an editable grid with 2 columns: start date and rate
the exchange rate screen removed from the code


packages/shared-domain/src/lib/date-helper.js
```diff
@@ -202,6 +202,19 @@ const formatTimeToApi = time => {
   })
 }
 
+// Format date as 'yyyymmdd'
+const formatDateToDayId = date => {
```
NoraBitarSM: naming the function 'formatDateToDayId' is very specific
what if you are using this function to format to yyyymmdd for another field ?

packages/repositories/src/repositories/MultiCurrencyRepository.js
```diff
@@ -29,9 +29,7 @@ export const MultiCurrencyRepository = {
   },
   ExchangeRates: {
     page: service + 'pageED',
```
NoraBitarSM: please check if pageED is unused also
since  the exchange rate page was removed from the code

packages/module-system/src/pages/exchange-tables/forms/ExchangeRatesGridForm.js
```diff
@@ -0,0 +1,135 @@
...
+export default function ExchangeRatesGridForm({ labels, maxAccess, store }) {
+  const { recordId } = store
+  const { getRequest, postRequest } = useContext(RequestsContext)
+  const { platformLabels } = useContext(ControlContext)
+
+  const invalidate = useInvalidate({
+    endpointId: MultiCurrencyRepository.ExchangeRates.page
+  })
+
+  const formatDate = dateStr => {
```
NoraBitarSM: add also this function in the date helper, so you can use from the helper the conversion both ways

NoraBitarSM:
not related but in the outer grid the currency column is always empty

PR #2252, opened 2026-06-04
AR-4744

imtithalhalalo:
point of sale maxEndTime optional


Ghinwa-Arab:
image
image

max end time, i saved with value 9:00 AM after refetching it returns 6:00 AM. it's normal?

Ghinwa-Arab (approved):

PR #2253, opened 2026-06-04
AR-4660

Ghinwa-Arab:
RV & PV: return the balance according to cash account not account reference
-> take into consideration when the cash account set by default (user's cash account or doc type cash account)


packages/shared-ui/src/components/Shared/Forms/FIReceiptVoucherForm.js
imtithalhalalo: || null

imtithalhalalo: || null

imtithalhalalo (approved):

PR #2263, opened 2026-06-08
AR-4764

imtithalhalalo:
add 2 optional combos: default Qty MU and default Unit Price MU
both MU combos filter by the MsId in brokerage/defaults measurement schedule field
these default fields should be filled in the fixing activities when selecting commodity pair


NoraBitarSM:
neg-deploy
fixing sale


i saved this record (FSA26SC-00001) with qty MU = gram
then in edit mode, remove the commodity pair, then add it again, the MU qty became 'test' because it's the default one, but baseQty was still 15 instead of 15 x 3 = 45 
Screenshot 2026-06-08 at 3 54 19 PM

NoraBitarSM (approved):

PR #2267, opened 2026-06-09
AR-4772

imtithalhalalo:
adding post button to leave request one day or more

### DIFF
--- a/packages/shared-ui/src/components/Shared/LeaveForm.js
+++ b/packages/shared-ui/src/components/Shared/LeaveForm.js
@@ -147,6 +147,7 @@ export const LeaveForm = ({ recordId, window }) => {
   }
 
   const isClosed = formik.values.wip == 2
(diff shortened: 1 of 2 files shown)


packages/shared-ui/src/components/Shared/LeaveForm.js
```diff
@@ -331,6 +343,18 @@ export const LeaveForm = ({ recordId, window }) => {
...
+      key: 'Locked',
+      condition: isPosted,
+      onClick: 'onUnpostConfirmation',
+      disabled: true
+    },
+    {
+      key: 'Unlocked',
+      condition: !isPosted,
+      onClick: onPost,
+      disabled: !isClosed
```
NoraBitarSM: post button should be enabled only when the record is released, it might need to wait for reviewers to approve it
here the button is becoming enabled based on the wip status

PR #2271, opened 2026-06-10
AR-4775

imtithalhalaloSM:
chart of account new fields
- 2 combo fields will be added
Group (1 - Balance sheet, 2 - Income Statement)
Section (1 - Assets, 2 - Liabilities, 3 - Equity, 4- Revenues, 5 - Expenses)
when i select Balance Sheet in Group combo, it will show Assets, Liabilities and Equity in second combo
when i select Income Statement in Group combo, it will show Revenues and Expenses in second
- add flName optional field


Ghinwa-Arab:
image

now we have 2 fields of same label, we need them like this? or we need to specify more the label of the added field?

Ghinwa-Arab (approved):

PR #2294, opened 2026-06-23
AR-4827

imtithalhalaloSM:
mf-prod-summary


NoraBitarSM:
Screenshot 2026-06-23 at 1 20 17 PM

1. the document type used in this form is not external, the reference should have been optional
2. increase the dropdown of the client lookup
3. i added a new record and posted it, when adding another new record, on import the same items were imported. Import should return only the items that were posted and not used in a summary

NoraBitarSM:
Screenshot 2026-06-24 at 3 20 20 PM

error on post

Screenshot 2026-06-24 at 3 26 35 PM

on each import i am getting ALL items again

packages/module-manufacturing/src/pages/mf-prod-summary/Forms/ProductionSummaryForm.js
```diff
@@ -0,0 +1,397 @@
...
+    invalidate()
+  }
+
+  async function onImportClick() {
+    const res = await getRequest({
+      extension: ManufacturingRepository.ProductionRequestItems.import,
+      parameters: ''
+    })
+
+    await postRequest({
```
NoraBitarSM: if no items were returned while importing, then no need to continue to the other requests
in this case, show a popup message that there is no items to import

packages/module-manufacturing/src/pages/mf-prod-summary/Forms/ProductionSummaryForm.js
```diff
@@ -0,0 +1,405 @@
...
+
+  async function onImportClick() {
+    const res = await getRequest({
+      extension: ManufacturingRepository.ProductionRequestItems.import,
+      parameters: ''
+    })
+
+    if (res?.list?.length > 0) {
+      await postRequest({
+        extension: ManufacturingRepository.ProductionSummary.set2,
```
NoraBitarSM, line 152: please show a saving toast message so the user can know that it was saved not only imported

NoraBitarSM (approved):

PR #2298, opened 2026-06-24
AR-4839

imtithalhalaloSM:
actors screen defaultId


NoraBitarSM:
this field represents the ID number (ra2em l hawiye) 
no thousands separator needed or right alignment

PR #2307, opened 2026-06-30
AR-4834

imtithalhalaloSM:
in Activities screens, we have an icon next to currency field, Change Rate, we need to add Security Access to who can or cannot open this screen using this icon


Ghinwa-Arab:
image
image

giving edit access only not enough to make the form editable? since it needs add access to do that

Ghinwa-Arab:
image

if i give the activity access to view only then i give the mcr form access to edit, it should be editable even if the main activity has access only for view right?

Ghinwa-Arab (approved):

PR #2308, opened 2026-06-30
AR-4848

imtithalhalaloSM:
adding a bit of spacing on the right of the numbers

### DIFF
--- a/packages/shared-ui/src/components/Shared/DataGrid/index.js
+++ b/packages/shared-ui/src/components/Shared/DataGrid/index.js
@@ -1351,6 +1351,14 @@ export function DataGrid({
           justify-content: flex-end !important;
         }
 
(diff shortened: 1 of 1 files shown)


packages/shared-ui/src/components/Shared/DataGrid/index.js
```diff
@@ -1351,6 +1351,14 @@ export function DataGrid({
           justify-content: flex-end !important;
         }
 
+        .agContainer:dir(ltr) :global(.ag-cell.numberCell) {
+          padding-right: 6px !important;
+        }
+
+        .agContainer:dir(rtl) :global(.ag-cell.numberCell) {
```
imtithalhalaloSM, line 1358: this was added because of the arabic

PR #2312, opened 2026-07-02
AR-4631

Ghinwa-Arab:
Item Details - qtyOnHand

now the api that returns qtyOnHand should filter according to the siteId if exists

### DIFF
--- a/packages/module-sales/src/pages/retail-trx/[functionId]/forms/RetailTransactionsForm.js
+++ b/packages/module-sales/src/pages/retail-trx/[functionId]/forms/RetailTransactionsForm.js
@@ -828,8 +828,11 @@ export default function RetailTransactionsForm({
         popup: row => stack({
           Component: ItemDetails,
           props: {
-            itemId: row?.itemId || null,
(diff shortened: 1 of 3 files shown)


NoraBitarSM:
1. where is the old qty field ? i asked for a new on hand per site field. we should have qty that returns the global availability (as it was) and a new field on hand per site
2. please show 0 if there is no on hand per site
3. please decrease the width of the qty and on hand per site fields and put them next to each other

packages/module-sales/src/pages/retail-trx/[functionId]/forms/RetailTransactionsForm.js
```diff
@@ -828,8 +828,10 @@ export default function RetailTransactionsForm({
         popup: row => stack({
           Component: ItemDetails,
           props: {
-            itemId: row?.itemId || null,
-            plId: formik.values?.header?.plId || null
+            values: {
+              itemId: row?.itemId || null,
+              plId: formik.values?.header?.plId || null
```
Comment (NoraBitarSM) [inline packages/module-sales/src/pages/retail-trx/[functionId]/forms/RetailTransactionsForm.js]: please send the siteId since there is one in the header

PR #2314, opened 2026-07-03
AR-4877

NoraBitarSM:
1. add picture column to datagrid - production order
you can hide the column from global auth

2. fixed qty condition to prevent saving with empty or 0 qty
3. removed Job Count column


Ghinwa-Arab:
image
try pressing tab to move to the second row. A new row is added correctly, but the focus doesn't move to any cell. If you keep pressing tab, the form start rendering on each click

Ghinwa-Arab:
image

there was an existing row, I added a new row and saved but after saving all the rows disappeared

Ghinwa-Arab:
image

it allows me to save with 0

PR #2328, opened 2026-07-10
AR-4896

imtithalhalaloSM:
we need Lock Record for these 2 screens
Draft Serial Invoice and Draft Serial Return screen - Sales


packages/module-sales/src/pages/sa-draft-serials-invoices/index.js
```diff
@@ -140,17 +143,47 @@ const DraftSerialsInvoices = () => {
   }
 
   const editDSI = obj => {
-    openForm(obj.recordId)
+    openForm(obj?.recordId, obj?.reference, obj?.status)
   }
 
-  async function openForm(recordId) {
+  async function openForm(recordId, reference, status) {
+    if (recordId && status !== 3) {
```
NoraBitarSM: the form is readonly also in the case of wip = completed 
no need to check for status posted

NoraBitarSM:
since both draft forms are shared, this means that we are calling them from at least 2 places
locking logic should be applied everywhere where we are opening the forms to prevent 2 people from editing the same record

NoraBitarSM:
Screenshot 2026-07-10 at 3 13 16 PM

on reopen it should lock the record
as i said lock should be requested to prevent 2 people from editing the same record

PR #2365, opened 2026-07-28
AR-4798 LIVE - APPROVED

imtithalhalaloSM:
material transfer is showing unsaved changes due to disableSkuLookup field
Screenshot 2026-07-27 at 10 42 55 PM


packages/shared-ui/src/components/Shared/Forms/MaterialsTransferForm.js
```diff
@@ -1016,6 +1016,7 @@ export default function MaterialsTransferForm({ recordId, window }) {
             ...res.record,
             plId,
             transfers: updatedTransfers,
+            disableSKULookup: formik.values.disableSKULookup || false,
```
Ghinwa-Arab: Are you sure that formik.values.disableSKULookup has already stored its value here? Since this useEffect runs on the initial render, it executes before the value is available

Ghinwa-Arab (approved):

PR #2371, opened 2026-07-29
AR-4965 - merged

imtithalhalaloSM:
min price issue
https://softmachine.atlassian.net/browse/AR-4965

### DIFF
--- a/packages/module-inventory/src/pages/iv-items/forms/SaleList.js
+++ b/packages/module-inventory/src/pages/iv-items/forms/SaleList.js
@@ -36,29 +36,36 @@ const SalesList = ({ store, labels, maxAccess, formikInitial }) => {
     ptName: isAnyFilled,
     value: isAnyFilled,
     vtName: isAnyFilled,
-    minPrice: row => (row.value > 0 || row?.plId > 0 || row?.ptName > 0 || row?.currencyId > 0) && row?.minPrice <= row?.value
+    minPrice: row => {
+      return {
+        required: false,
+        invalid: row.minPrice > row.value
+      }
+    }
   }
   const { schema, requiredFields } = createConditionalSchema(conditions, true, maxAccess, 'items')
 
+  const initialValues = {
+    defSaleMUId: store.measurementId || '',
+    pgId: store.priceGroupId || '',
+    returnPolicyId: store.returnPolicy || '',
+    items: [
+      {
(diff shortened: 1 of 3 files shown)


packages/module-inventory/src/pages/iv-items/forms/SaleList.js
```diff
@@ -22,47 +22,51 @@ const SalesList = ({ store, labels, maxAccess, formikInitial }) => {
...
-    vtName: isAnyFilled,
-    minPrice: row => (row.value > 0 || row?.plId > 0 || row?.ptName > 0 || row?.currencyId > 0) && row?.minPrice <= row?.value
+    currencyId: row => row?.currencyId,
+    plId: row => row?.plId,
+    ptName: row => row?.ptName,
+    value: row => row?.value || row?.value === 0,
+    vtName: row => row?.vtName,
+    minPrice: row => {
+      if (row?.minPrice == null || row?.minPrice === '') {
+        return { valid: true, optional: true }
```
Ghinwa-Arab, line 33: do we really need this feature to be added, u cannot concatenate to the schema the condition u need & add nullable?
so when no value it's treated as optional and when u have value it will check the condition

ex:
```
const salesRowSchema = schema.concat(
  yup.object({
    minPrice: yup.number().nullable()
    .when('value', {
        is: value => value !== null && value !== undefined && value !== '',
        then: schema => schema.required()
      })
      .test(function () {
        **here add the condition u need**
      })
  })
)
```
and this pass this updated schema to the formik

packages/shared-domain/src/lib/validation.js
```diff
@@ -10,30 +10,43 @@ function conditionalField(fieldValidators, fieldKey, allowNoLines) {
     const allRows = this.options.context?.[arrayKey] || []
     if (allowNoLines || (!allowNoLines && allRows.length > 1)) {
       const isAnyFieldFilled = Object.entries(fieldValidators).some(([, fn]) => {
-        return !!fn(row)
+        const res = fn(row)
+        if (typeof res === 'object' && res?.optional) return false
```
Ghinwa-Arab, line 14: if concatenated schema works then here u will modify the code to handle global auth access only, u will not check res?.optional anymore

Ghinwa-Arab (approved):

PR #2383, opened 2026-08-03
AR-4920

imtithalhalaloSM:
adding reference
Dimension Group screen - add Reference max 10 chars not required
delete issue reported to hasan


Ghinwa-Arab:
image

on delete

PR #2398, opened 2026-08-07
AR-4967 nora

NoraBitarSM:
column resize while keeping the freeze feature


Ghinwa-Arab:
image

The displayed columns shrink after I resize the machineName column and hide it from global auth.

Ghinwa-Arab:
image

After resizing & hiding the machineName column if I resize the notes column and then show the machineName column again, it appears at the end of the table and its resized width is lost

PR #2404, opened 2026-08-11
AR-4953

Ghinwa-Arab:
add filters to processed shift punches


packages/module-hr/src/pages/hr-processed-punches/index.js
imtithalhalaloSM, line 120: missing name

imtithalhalaloSM (approved):

PR #2405, opened 2026-08-11
AR-5014

NoraBitarSM:
https://softmachine.atlassian.net/browse/AR-5014


Ghinwa-Arab:
image

according to the task screenshot, commodity column is replaced by currency & qty columns + release status column is missing

Ghinwa-Arab:
table row coloring is missing

packages/module-bt/src/pages/bt-event-order-inquiry/index.js
```diff
@@ -0,0 +1,159 @@
...
+    filter: {
+      filterFn: fetchWithFilter
+    }
+  })
+
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      refetch()
+    }, 60000)
```
Ghinwa-Arab: it's refetching every 1 min not 5 mins

Ghinwa-Arab (changes requested):

PR #2408, opened 2026-08-11
AR-5001

imtithalhalaloSM:
getPackRL changes now it only returns layouts (already combined) and defaultLayoutId
in delivery-settings we still need combination between reportTemplates and layouts (from getPackRL2)


Ghinwa-Arab:
image

for this activity I have one active custom report, we don't want to show it?? its showing report layout reports only

Ghinwa-Arab:
image

When I assign all report layout as inactive, the custom one is displayed. However, when I reactivate all report layout accounts, the api no longer returns any of them

Ghinwa-Arab:
image
image
i detect the issue again in sales return, when i deactivate all reports in report layouts then activate them back api not returning them

Ghinwa-Arab (approved):

PR #2411, opened 2026-08-12
Gold price ticker

NoraBitarSM:


packages/shared-ui/src/components/Shared/GoldPriceTicker/index.js
```diff
@@ -0,0 +1,221 @@
...
+  const data = await res.json()
+
+  return data.price 
+}
+
+async function fetchSpotInCurrency(symbol, currency, rate) {
+  if (currency === 'USD') return fetchSpotUSD(symbol)
+
+  if (rate) {
+    const usd = await fetchSpotUSD(symbol)
```
imtithalhalaloSM, line 20: it is better to check if there is number returned before the calculation

packages/shared-ui/src/components/Shared/GoldPriceTicker/index.js
```diff
@@ -0,0 +1,221 @@
...
+      try {
+        const needsSeparateUsdFetch = karatCurrency !== 'USD'
+
+        const [xauKarat, xagUsd, xauUsd] = await Promise.all([
+          fetchSpotInCurrency('XAU', karatCurrency, rate),
+          fetchSpotUSD('XAG'),
+          needsSeparateUsdFetch ? fetchSpotUSD('XAU') : Promise.resolve(null)
+        ])
+
+        const goldPerGramKarat = xauKarat / GRAMS_PER_OZ
```
imtithalhalaloSM, line 50: also here to check if  xauKarat is number

PR #2414, opened 2026-08-13
AR-5007

imtithalhalaloSM:
account balance fields
https://softmachine.atlassian.net/browse/AR-5007


packages/module-financials/src/pages/fi-accounts/forms/AccountBalanceTable.js
```diff
@@ -80,6 +114,29 @@ const AccountBalanceForm = ({ labels, maxAccess, store }) => {
             pagination={false}
           />
         </Grow>
+        <Fixed>
+          <Grid container spacing={2} padding={2}>
+            <Grid item xs={6}></Grid>
```
Ghinwa-Arab: u don't need to add empty grid, u can add to the container flex end or something like this

Ghinwa-Arab (approved):

PR #2429, opened 2026-08-18
AR-4967Live

Ghinwa-Arab:
column resize while keeping the freeze feature


imtithalhalaloSM:
when i do refresh for the screen after resizing it is returning the columns as it was and then resize it directly

packages/shared-ui/src/components/Shared/Table/index.js
```diff
@@ -1088,91 +1199,47 @@ const Table = ({
...
+        applyOrder: false
+      }
+    )
 
     if (!pinned) {
       const targetIndex = originalOrder.indexOf(colId)
-      
       if (targetIndex > -1) {
-        gridApiRef.current?.columnApi?.moveColumn(colId, targetIndex)
+        gridApiRef.current.columnApi.moveColumn(colId, targetIndex)
```
imtithalhalaloSM: why did you remove the checks

Ghinwa-Arab: no need since at the top of the function we are checking if gridApiRef.current?.columnApi don't exists then return

packages/shared-ui/src/components/Shared/Table/index.js
```diff
@@ -1088,91 +1199,47 @@ const Table = ({
 
     const originalOrder = columnDefs.map(col => col.field)
 
-    gridApiRef.current?.columnApi?.applyColumnState({
-      state: [{ colId, pinned }],
-      applyOrder: false
-    })
+    gridApiRef.current.columnApi.applyColumnState(
```
imtithalhalaloSM: same here

imtithalhalaloSM (approved):

